PR #1, opened 2019-03-31
Attempt to parse project from .tfplan before asking for --project

nstogner:
The terraform plan contains an un-interpolated provider configuration. In this PR, we attempt to pull a hardcoded `project` field for the `google` provider rather than requiring it to be passed in via the flag `--project`. This PR does not attempt to run interpolations.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -27,11 +27,9 @@ func init() {
 
 	validateCmd.Flags().StringVar(&flags.validate.policyPath, "policy-path", "", "Path to directory containing validation policies")
 	validateCmd.MarkFlagRequired("policy-path")
-	validateCmd.Flags().StringVar(&flags.validate.project, "project", "", "Provider project")
-	validateCmd.MarkFlagRequired("project")
+	validateCmd.Flags().StringVar(&flags.validate.project, "project", "", "Provider project override")
 
-	convertCmd.Flags().StringVar(&flags.convert.project, "project", "", "Provider project")
-	convertCmd.MarkFlagRequired("project")
+	convertCmd.Flags().StringVar(&flags.convert.project, "project", "", "Provider project override")
 
 	validateCmd.Flags().BoolVar(&flags.validate.outputJSON, "output-json", false, "Print violations as JSON")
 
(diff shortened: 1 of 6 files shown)


tfgcv/planned_assets.go
```diff
@@ -59,3 +64,30 @@ func ReadPlannedAssets(path, project string) ([]google.Asset, error) {
...
+			// If the provider has a hardcoded project string, return it.
+			if !strings.Contains(prj, "${") {
+				return prj, nil
+			}
+
+			return "", ErrParsingProviderProject
+		}
+	}
+
+	return "", nil
```
t12g: Why do we return nil for error instead of ErrParsingProviderProject?

cmd/root.go
```diff
@@ -27,11 +27,9 @@ func init() {
 
 	validateCmd.Flags().StringVar(&flags.validate.policyPath, "policy-path", "", "Path to directory containing validation policies")
 	validateCmd.MarkFlagRequired("policy-path")
-	validateCmd.Flags().StringVar(&flags.validate.project, "project", "", "Provider project")
-	validateCmd.MarkFlagRequired("project")
+	validateCmd.Flags().StringVar(&flags.validate.project, "project", "", "Provider project override")
```
t12g: nit: can we expand on the override behavior a bit in the flag documentation?

tfgcv/planned_assets.go
nstogner: Provider `project` field is optional.

nstogner: ... It is not an error in terraform if the `project` field is not specified. We are just error-ing out here if we are unable to parse it.

cmd/root.go
nstogner: Updated

tfgcv/planned_assets.go
t12g: Right, so I was curious about the difference in behavior. In both scenarios, we don't have a project id that we could use.
We return ErrParsingProviderProject when the google config is present and has a project id that is not hardcoded.
We return nil when the google config is not present at all. So if the --project flag is not set and there's no google config, we'll continue in line 50 to create a new Converter with an empty project string. I was a bit confused by that. Can we add a comment?

nstogner: Will do

Issue #21, opened 2019-04-29
Overriding ancestry discovery


yukinying:
Ancestry discovery (https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/converters/google/convert.go#L220) use Google APIs to retrieve the ancestry path. In scenario that validation could be run in an isolated environment, it would be desirable to provide an option to suppress such behavior. It would be great if the tool can use a flag like --folder (as in --project) such that it can provide the information needed instead of dynamic ancestry discovery.

morgante:
Fixed by #22.

PR #22, opened 2019-04-30
add --folder flag to override ancestry discovery

yukinying:
The idea is to pre populate ancestry cache if `--folder` is provided from the command line. This should fix issue #21.

### DIFF
--- a/cmd/convert.go
+++ b/cmd/convert.go
@@ -36,7 +36,8 @@ Note:
   resources.
 
 Example:
-  terraform-validator convert ./example/terraform.tfplan --project my-project
+  terraform-validator convert ./example/terraform.tfplan --project my-project \
+    --ancestry organization/my-org/folder/my-folder
 `,
 	PreRunE: func(c *cobra.Command, args []string) error {
 		if len(args) != 1 {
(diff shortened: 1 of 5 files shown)


cmd/convert.go
```diff
@@ -36,7 +36,8 @@ Note:
   resources.
 
 Example:
-  terraform-validator convert ./example/terraform.tfplan --project my-project
+  terraform-validator convert ./example/terraform.tfplan --project my-project \
+    --folder organization/my-org/project/my-project
```
morgante: Instead of passing the project, I think folder should just include the parent. ex. `organization/my-org` or `organization/my-org/folder/my-folder`.

Also let's call the flag `--ancestry` instead of `--folder`.

cmd/convert.go
```diff
@@ -36,7 +36,8 @@ Note:
   resources.
 
 Example:
-  terraform-validator convert ./example/terraform.tfplan --project my-project
+  terraform-validator convert ./example/terraform.tfplan --project my-project \
+    --ancestry organization/my-org/folder/my-folder/project/my-project
```
morgante: @yukinying Specifying the project in both `--project` and `--ancestry` seems redundant. Ancestry should just be the parent path (`organization/my-org/folder/my-folder`).

yukinying: Agree. This is redundant. Code updated to add project to ancestry.

PR #28, opened 2019-05-20
Start sending Terraform Validator user agent string for usage tracking

t12g:
The user agent string includes build version, so we are embedding the version string as part of build process.

Also add a new "version" command to display the version and rev the terraform-google-conversion version to pick up the latest resource manager client that includes client option support.

The current version string is just the date (`date +%Y-%m-%d`). That's problematic if we ever need to have 2 releases in the same day, but we can solve that problem separately.


morgante:
@t12g It looks like this also includes a lot of unrelated resource changes; are those needed for this PR?

t12g:
@morgante Do you mean the changes under /vendor? They're brought in as the result of updating terraform-google-conversion, and a new version of terraform-google-conversion is needed to specify the cloud resource manager client options (they aren't available in the current version).

t12g:
ping?

PR #37, opened 2019-06-06
Simplify integration tests

morgante:
The current integration tests are tightly coupled to the underlying API libraries and casts the data to structs.

This replaces them with an approach built on simple JSON files, so adding new tests is just a matter of defining equivalent Terraform and expected CAI output—without involving the implementation details of the API libraries.


test/convert_test.go
```diff
@@ -54,30 +53,46 @@ func TestConvert(t *testing.T) {
...
+
+		jsonFixtures[fixtureName] = fixtureData
+	}
+
 	t.Run("Disk", func(t *testing.T) {
-		requireEqualJSONValues(t,
-			// Expected:
-			data.Disk,
-			// Received:
+		requireEqualJSON(t,
```
t12g: since the test cases now only differ in fixture name and asset type, can we change this into a table driven test instead of having to explicitly call t.Run() on each case?

test/convert_test.go
```diff
@@ -54,30 +53,46 @@ func TestConvert(t *testing.T) {
...
+	jsonFixtures := make(map[string][]byte)
+
+	matches, _ := filepath.Glob(filepath.Join(jsonGenerateDir, "*.json"))
+	for _, fixturePath := range matches {
+		fixtureFileName := strings.TrimPrefix(fixturePath, jsonGenerateDir+"/")
+		fixtureName := strings.TrimSuffix(fixtureFileName, ".json")
+
+		fixtureData, err := ioutil.ReadFile(fixturePath)
+		if err != nil {
+			t.Fatalf("Reading: %v", err)
```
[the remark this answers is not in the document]
morgante: Fixed in bc67aa68011d8b43eb6db378de8c52fd1e208baf

test/convert_test.go
morgante: Good idea, implemented in 3c1fd17d3e678f24e9a4d21a98e7755efad7d5e4

test/convert_test.go
```diff
@@ -54,30 +63,28 @@ func TestConvert(t *testing.T) {
...
-			data.ProjectBillingInfo,
-			assetsByType["cloudbilling.googleapis.com/ProjectBillingInfo"][0].Resource.Data,
-			&cloudbillingv1.ProjectBillingInfo{},
-		)
-	})
+		jsonFixtures[fixtureName] = fixtureData
+	}
+
+	for _, tt := range conversionTests {
+		// actual := assetsByType[tt.assetType][0].Resource.Data
```
t12g: delete?

morgante: Thanks, good catch.

PR #38, opened 2019-06-06
add offline flag

yukinying:
The setup of resource manager client relies on Google OAuth library which would read the ADC json file and acquire a token, even if the client is not used in subsequent logic. I propose adding an offline flag so the client would not be setup for offline use case.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -29,9 +29,11 @@ func init() {
 	validateCmd.MarkFlagRequired("policy-path")
 	validateCmd.Flags().StringVar(&flags.validate.project, "project", "", "Provider project override (override the default project configuration assigned to the google terraform provider when validating resources)")
 	validateCmd.Flags().StringVar(&flags.validate.ancestry, "ancestry", "", "Override the ancestry location of the project when validating resources")
+	validateCmd.Flags().BoolVar(&flags.validate.offline, "offline", false, "Do not connect to GCP API")
 
 	convertCmd.Flags().StringVar(&flags.convert.project, "project", "", "Provider project override (override the default project configuration assigned to the google terraform provider when converting resources)")
 	convertCmd.Flags().StringVar(&flags.convert.ancestry, "ancestry", "", "Override the ancestry location of the project when validating resources")
+	convertCmd.Flags().BoolVar(&flags.convert.offline, "offline", false, "Do not connect to GCP API")
 
 	validateCmd.Flags().BoolVar(&flags.validate.outputJSON, "output-json", false, "Print violations as JSON")
 
@@ -52,10 +54,12 @@ var flags struct {
 	convert struct {
 		project  string
 		ancestry string
+		offline  bool
 	}
 	validate struct {
 		project    string
 		ancestry   string
+		offline    bool
 		policyPath string
 		outputJSON bool
 	}
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -16,6 +16,7 @@ package google
 
 import (
 	"fmt"
+	"log"
 	"sort"
 	"strings"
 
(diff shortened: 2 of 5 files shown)


converters/google/convert.go
```diff
@@ -249,10 +249,15 @@ func (c *Converter) augmentAsset(tfData converter.TerraformResourceData, cfg *co
 // projects. It implements a cache because many resources share the same
 // project.
 func (c *Converter) getAncestry(project string) (string, error) {
+	fmt.Printf("getAncestry: project=%s\n", project)
```
morgante: Can we hide this behind a verbose flag?

cmd/root.go
```diff
@@ -29,9 +29,11 @@ func init() {
 	validateCmd.MarkFlagRequired("policy-path")
 	validateCmd.Flags().StringVar(&flags.validate.project, "project", "", "Provider project override (override the default project configuration assigned to the google terraform provider when validating resources)")
 	validateCmd.Flags().StringVar(&flags.validate.ancestry, "ancestry", "", "Override the ancestry location of the project when validating resources")
+	validateCmd.Flags().BoolVar(&flags.validate.offline, "offline", false, "Do not connect to GCP API")
 
 	convertCmd.Flags().StringVar(&flags.convert.project, "project", "", "Provider project override (override the default project configuration assigned to the google terraform provider when converting resources)")
 	convertCmd.Flags().StringVar(&flags.convert.ancestry, "ancestry", "", "Override the ancestry location of the project when validating resources")
+	convertCmd.Flags().BoolVar(&flags.convert.offline, "offline", false, "Do not connect to GCP API")
```
morgante: Can we add a check that ancestry must be provided if offline is set?

converters/google/convert.go
yukinying: yes. Changed to log.Printf which only logs with --verbose.

cmd/root.go
yukinying: Done.

converters/google/convert.go
```diff
@@ -249,10 +250,15 @@ func (c *Converter) augmentAsset(tfData converter.TerraformResourceData, cfg *co
 // projects. It implements a cache because many resources share the same
 // project.
 func (c *Converter) getAncestry(project string) (string, error) {
+	log.Printf("getAncestry: project=%s", project)
```
t12g: won't this be too spammy? can we only print this when `--verbose` is set?

t12g:
We also have similar usage in https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/converters/google/convert.go#L94. The`LoadAndValidate()` func does the same thing. Actually looking at it again, do we actually need it? It doesn't seem to do anything other than initializaing a bunch of clients.

converters/google/convert.go
```diff
@@ -249,10 +250,15 @@ func (c *Converter) augmentAsset(tfData converter.TerraformResourceData, cfg *co
...
 	if path, ok := c.ancestryCache[project]; ok {
 		return path, nil
 	}
 
+	if c.resourceManager == nil {
```
t12g: Since we allow nil resourceManager now, can we document this in `NewConverter()` and the `Converter` struct?

yukinying:
> We also have similar usage in https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/converters/google/convert.go#L94. The`LoadAndValidate()` func does the same thing. Actually looking at it again, do we actually need it? It doesn't seem to do anything other than initializaing a bunch of clients.

Right, I missed that.

yukinying:
I have updated `NewConverter` to take `offline` as an argument.

converters/google/convert.go
```diff
@@ -86,13 +87,16 @@ type AssetResource struct {
...
-func NewConverter(resourceManager *cloudresourcemanager.Service, project, ancestry, credentials string) (*Converter, error) {
+func NewConverter(resourceManager *cloudresourcemanager.Service, project, ancestry, credentials string, offline bool) (*Converter, error) {
 	cfg := &converter.Config{
 		Project:     project,
 		Credentials: credentials,
 	}
-	if err := cfg.LoadAndValidate(); err != nil {
-		return nil, errors.Wrap(err, "configuring")
+
+	if !offline {
```
t12g: have you tried removing `LoadAndValidate()` invocation altogether? Since we don't expect any API calls to be made, it doesn't seem to make sense to initialize those clients.

yukinying: I created another pull request at #39 for removing `LoadAndValidate()`.

converters/google/convert.go
yukinying: There is a log suppression logic in root.go when --verbose is set:

```
		if !flags.verbose {
			// Suppress chatty packages.
			log.SetOutput(ioutil.Discard)
		}
```

yukinying:
Please allow me to add integration test before this gets accepted.

yukinying:
This is pending on #40 for having integration test structure on `validate` command.

morgante:
@yukinying Can you also make sure to add an integration test which exercises the offline flag before merging?

morgante:
@yukinying Please resolve conflicts if you'd like to see this picked up.

PR #39, opened 2019-06-07
remove calling of cfg.LoadAndValidate

yukinying:
Based on discussion in #38, there is no need to call LoadAndValidate. In particular, the function perform OAuth dance which creates extra network dependencies.

### DIFF
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -91,9 +91,6 @@ func NewConverter(resourceManager *cloudresourcemanager.Service, project, ancest
 		Project:     project,
 		Credentials: credentials,
 	}
-	if err := cfg.LoadAndValidate(); err != nil {
-		return nil, errors.Wrap(err, "configuring")
-	}
 
 	ancestryCache := make(map[string]string)
 	if ancestry != "" {


yukinying:
I don't have any integration test cases that can be used to verify this. After I looked at the code more deeply, it looks like `converter` is using the `cfg` with a function map. This dynamic function loading behavior would probably be hard to scope. I would like to close this pull request as I am not having the confidence here.

morgante:
Yeah, at a minimum we should make sure that the integration tests pass (`make test-integration`) before merging a change like this.

t12g:
However, if this PR breaks anything, that means the same breakage exists in offline mode as well.
Here's my suggestion - let's get the offline PR in first. Let it soak a bit. If everything works as expected, then we can remove LoadAndValidate() altogether. Does that sound reasonable?

yukinying:
> Yeah, at a minimum we should make sure that the integration tests pass (`make test-integration`) before merging a change like this.

It looks like `make test-integration` does not work in my local development environment:

```
=== RUN   TestConvert
--- FAIL: TestConvert (1.69s)
    convert_test.go:53: fork/exec ../bin/terraform-validator: no such file or directory:
FAIL
FAIL	github.com/GoogleCloudPlatform/terraform-validator/test	1.898s
Makefile:15: recipe for target 'test-integration' failed
make: *** [test-integration] Error 1
```

It may be the case I messed up something. Just pasting here as an FYI.

yukinying:
> However, if this PR breaks anything, that means the same breakage exists in offline mode as well.
> Here's my suggestion - let's get the offline PR in first. Let it soak a bit. If everything works as expected, then we can remove LoadAndValidate() altogether. Does that sound reasonable?

SGTM

morgante:
> It looks like make test-integration does not work in my local development environment:

You need to have a binary build (`make build`) before running the tests.

PR #40, opened 2019-06-07
add integration test for `validate` command.

yukinying:
Note that the added logic needs to be run inside `TestConvert` as it needs the generated `test.tfplan` file.

### DIFF
--- a/test/convert_test.go
+++ b/test/convert_test.go
@@ -15,16 +15,18 @@
 package test
 
 import (
-	"bytes"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
-	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"testing"
 
 	"github.com/GoogleCloudPlatform/terraform-validator/converters/google"
 )
+
+const samplePolicyPath = "sample_policies"
 
 var conversionTests = []struct {
 	name      string
@@ -40,21 +42,19 @@ var conversionTests = []struct {
 func TestConvert(t *testing.T) {
 	_, cfg := setup(t)
 
-	cmd := exec.Command(filepath.Join("..", "bin", "terraform-validator"),
+	err, stdOutput, errOutput := runWithCred(t, cfg.credentials,
+		filepath.Join("..", "bin", "terraform-validator"),
 		"convert",
 		"--project", cfg.project,
 		planPath,
 	)
-	cmd.Env = []string{"GOOGLE_APPLICATION_CREDENTIALS=" + cfg.credentials}
-	var stderr, stdout bytes.Buffer
-	cmd.Stderr, cmd.Stdout = &stderr, &stdout
 
-	if err := cmd.Run(); err != nil {
-		t.Fatalf("%v:\n%v", err, stderr.String())
+	if err != nil {
+		t.Fatalf("%v:\n%v", err, errOutput)
 	}
 
 	var assets []google.Asset
-	if err := json.Unmarshal(stdout.Bytes(), &assets); err != nil {
+	if err := json.Unmarshal(stdOutput, &assets); err != nil {
 		t.Fatalf("unmarshaling: %v", err)
 	}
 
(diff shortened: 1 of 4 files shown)


test/convert_test.go
```diff
@@ -86,4 +84,32 @@ func TestConvert(t *testing.T) {
...
 		})
 	}
+
+	validationTests := []struct {
+		desc            string
+		policyPath      string
+		wantError       bool
+		wantOutputRegex string
+	}{
+		{"always_violate", "sample_policies/always_violate", true, "Constraint always_violates_all on resource"},
```
t12g: nit: can we use the `field: value` form for the struct literal? it uses up more vertical space, but I find it easier to follow along

test/sample_policies/always_violate/policies/constraints/always_violates.yaml
```diff
@@ -0,0 +1,25 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+apiVersion: constraints.gatekeeper.sh/v1alpha1
+kind: GCPAlwaysViolatesConstraintV1
+metadata:
+  name: always_violates_all
+spec:
+  constraintVersion: 0.1.0
```
t12g: let's delete this version field; it's no longer used

test/sample_policies/always_violate/lib/constraints.rego
```diff
@@ -0,0 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+package validator.gcp.lib
+
+# Function to fetch the constraint spec
+# Usage:
+# get_constraint_params(constraint, params)
```
t12g: can we inline these in gcp_always_violates_v1.yaml?

test/sample_policies/always_violate/policies/templates/gcp_always_violates_v1.yaml
```diff
@@ -0,0 +1,65 @@
...
+  crd:
+    spec:
+      names:
+        kind: GCPAlwaysViolatesConstraintV1
+        plural: gcpalwaysviolatesconstraintsv1
+      validation:
+        openAPIV3Schema:
+          properties: {}
+  targets:
+   validation.gcp.forsetisecurity.org:
```
t12g: can we use the new form of targets here?
example: https://github.com/forseti-security/config-validator/blob/10863ea54f95e93d392fd7849f7148ec668b6197/pkg/gcv/configs/config_test.go#L220

We're trying to retire this old format.

yukinying:
Code updated. PTAL @t12g .

test/sample_policies/always_violate/policies/templates/gcp_always_violates_v1.yaml
t12g: I don't see `- target: validation.gcp.forsetisecurity.org` in the latest snapshot?

morgante: Instead of manually updating this, could we simply use a subtree of the policy library?

yukinying: Yes and no. I would prefer having a clean set of tests that does not depends on policy library, and then another set of tests that depends on it. So this pull request falls into the former.

yukinying: > I don't see - target: validation.gcp.forsetisecurity.org in the latest snapshot?

Sorry I was trying to put comments but marked the conversation resolved. I am not able to understand the new syntax. There seems to be an YAML syntax issue there in the following snippet, in the third line (or did I miss something)?

```
  targets:
    - target:validation.gcp.forsetisecurity.org
      rego: |
            # Some random
            # rego code
```

yukinying: Resolved. It looks like the referenced code in https://github.com/forseti-security/config-validator/blob/master/pkg/gcv/configs/config_test.go missed a space between `target:` and `validation.gcp.forsetisecurity.org`.

yukinying: I opened https://github.com/forseti-security/config-validator/pull/43 to fix the yaml syntax error in config-validator.

PR #45, opened 2019-06-18
Update config validator library

t12g:
To pick up https://github.com/forseti-security/config-validator/pull/45


t12g:
Should be good now. PTAL.

PR #49, opened 2019-06-28
integration test without dependencies on shelling out terraform

yukinying:
This add integration test without using terraform binary, and will potentially facilitate testing of offline capability of the validator. The terraform plan structure are taken from the sample in Terraform Google Cloud Platform provider from https://www.terraform.io/docs/providers/google/.

### DIFF
--- a/test/utils.go
+++ b/test/utils.go
@@ -63,6 +63,7 @@ func setup(t *testing.T) (data, config) {
 type config struct {
 	project     string
 	credentials string
+	policy      string
 }
 
 func configure(t *testing.T) config {
@@ -78,6 +79,12 @@ func configure(t *testing.T) config {
 	if !ok {
 		t.Fatal("missing required env var TEST_CREDENTIALS")
 	}
+
+	cfg.policy, ok = os.LookupEnv("TEST_POLICY")
+	if !ok {
+		t.Fatal("missing required env var TEST_POLICY to specify the path to the policy library")
+	}
+
 	// Make credentials path relative to repo root rather than
 	// test/ dir if it is a relative path.
 	if !filepath.IsAbs(cfg.credentials) {
--- a/test/validate_test.go
+++ b/test/validate_test.go
@@ -1,0 +1,294 @@
+package test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/terraform-validator/converters/google"
+	"github.com/GoogleCloudPlatform/terraform-validator/tfgcv"
+	"github.com/GoogleCloudPlatform/terraform-validator/tfplan"
+	"github.com/hashicorp/terraform/terraform"
+	"google.golang.org/api/cloudresourcemanager/v1"
+	"google.golang.org/api/option"
+)
+
+func TestValidate(t *testing.T) {
+	// Retrieving test configurations needed for building test cases.
+	cfg := configure(t)
+
+	// The test case takes a diff in the form of terraform.instanceDiff as an
+	// input, and compares number of violations returned with wantViolationCount.
+	cases := []struct {
+		kind               string
+		name               string
+		diff               *terraform.InstanceDiff
+		wantViolationCount int
+	}{
+		{
+			kind: "google_organization_iam_policy",
+			name: "foo_organization_iam_policy",
+			diff: &terraform.InstanceDiff{
+				Attributes: map[string]*terraform.ResourceAttrDiff{
+					"org_id":      &terraform.ResourceAttrDiff{New: "12345678"},
+					"policy_data": &terraform.ResourceAttrDiff{New: `{"bindings":[{"members":["user:jane@example.com"],"role":"roles/editor"}]}`},
+				},
+			},
+			wantViolationCount: 2,
+		},
+		{
+			kind: "google_organization_iam_binding",
+			name: "foo_organization_iam_binding",
+			diff: &terraform.InstanceDiff{
+				Attributes: map[string]*terraform.ResourceAttrDiff{
+					"members.#":        &terraform.ResourceAttrDiff{New: "1"},
+					"members.12345678": &terraform.ResourceAttrDiff{New: "alice@gmail.com"},
+					"org_id":           &terraform.ResourceAttrDiff{New: "12345678"},
+					"role":             &terraform.ResourceAttrDiff{New: "role/browser"},
+				},
+			},
+			wantViolationCount: 2,
+		},
+		{
+			kind: "google_organization_iam_member",
+			name: "foo_organization_iam_member",
+			diff: &terraform.InstanceDiff{
+				Attributes: map[string]*terraform.ResourceAttrDiff{
+					"member": &terraform.ResourceAttrDiff{New: "alice@gmail.com"},
+					"org_id": &terraform.ResourceAttrDiff{New: "12345678"},
+					"role":   &terraform.ResourceAttrDiff{New: "role/browser"},
+				},
+			},
+			wantViolationCount: 2,
+		},
+		{
+			kind: "google_folder_iam_policy",
+			name: "foo_folder_iam_policy",
+			diff: &terraform.InstanceDiff{
+				Attributes: map[string]*terraform.ResourceAttrDiff{
+					"folder":      &terraform.ResourceAttrDiff{New: "${google_folder.department1.name}"},
+					"policy_data": &terraform.ResourceAttrDiff{New: `{"bindings":[{"members":["user:jane@example.com"],"role":"roles/editor"}]}`},
+				},
+			},
+			wantViolationCount: 2,
+		},
+		{
+			kind: "google_folder_iam_binding",
+			name: "foo_folder_iam_binding",
+			diff: &terraform.InstanceDiff{
+				Attributes: map[string]*terraform.ResourceAttrDiff{
+					"members.#":        &terraform.ResourceAttrDiff{New: "1"},
+					"members.12345678": &terraform.ResourceAttrDiff{New: "alice@gmail.com"},
+					"folder":           &terraform.ResourceAttrDiff{New: "${google_folder.department1.name}"},
+					"role":             &terraform.ResourceAttrDiff{New: "role/browser"},
+				},
+			},
+			wantViolationCount: 2,
+		},
+		{
+			kind: "google_folder_iam_member",
+			name: "foo_folder_iam_member",
+			diff: &terraform.InstanceDiff{
+				Attributes: map[string]*terraform.ResourceAttrDiff{
+					"member": &terraform.ResourceAttrDiff{New: "alice@gmail.com"},
+					"folder": &terraform.ResourceAttrDiff{New: "${google_folder.department1.name}"},
+					"role":   &terraform.ResourceAttrDiff{New: "role/browser"},
+				},
+			},
+			wantViolationCount: 2,
+		},
+		{
+			kind: "google_project_iam_policy",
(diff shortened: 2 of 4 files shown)


yukinying:
@morgante @t12g PTAL.

test/utils.go
```diff
@@ -78,6 +79,12 @@ func configure(t *testing.T) config {
 	if !ok {
 		t.Fatal("missing required env var TEST_CREDENTIALS")
 	}
+
+	cfg.policy, ok = os.LookupEnv("TEST_POLICY")
+	if !ok {
+		t.Fatal("missing required env var TEST_POLICY")
```
t12g: Can we add a description of what this variable should look like?

test/validate_test.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil {
+		t.Fatalf("constructing resource manager client: %v", err)
+	}
+	baseResourceInstanceDiff := &terraform.InstanceDiff{
+		Attributes: map[string]*terraform.ResourceAttrDiff{
+			"name": &terraform.ResourceAttrDiff{New: "new-foo"},
+		},
+	}
+
+	type testcase struct {
```
t12g: nit: just inline this struct type like https://github.com/golang/go/wiki/TableDrivenTests#example-of-a-table-driven-test?

test/validate_test.go
```diff
@@ -0,0 +1,277 @@
...
+)
+
+func TestValidate(t *testing.T) {
+	cfg := configure(t)
+	ancestry := "organization/example.com/folder/foobar"
+	resourceManager, err := cloudresourcemanager.NewService(context.Background(), option.WithoutAuthentication())
+	if err != nil {
+		t.Fatalf("constructing resource manager client: %v", err)
+	}
+	baseResourceInstanceDiff := &terraform.InstanceDiff{
```
t12g: I'm not familiar with the terraform diff business... can we add a comment to explain why this is needed?

test/validate_test.go
```diff
@@ -0,0 +1,277 @@
...
+	"github.com/GoogleCloudPlatform/terraform-validator/tfgcv"
+	"github.com/GoogleCloudPlatform/terraform-validator/tfplan"
+	"github.com/hashicorp/terraform/terraform"
+	"google.golang.org/api/cloudresourcemanager/v1"
+	"google.golang.org/api/option"
+)
+
+func TestValidate(t *testing.T) {
+	cfg := configure(t)
+	ancestry := "organization/example.com/folder/foobar"
```
t12g: nit: can we define ancestry and resource manager closer to the point of usage?

test/validate_test.go
```diff
@@ -0,0 +1,277 @@
...
+					"settings.0.disk_autoresize":  &terraform.ResourceAttrDiff{New: "true"},
+					"settings.0.pricing_plan":     &terraform.ResourceAttrDiff{New: "PER_USE"},
+					"settings.0.replication_type": &terraform.ResourceAttrDiff{New: "SYNCHRONOUS"},
+					"settings.0.tier":             &terraform.ResourceAttrDiff{New: "db-f1-micro"},
+				},
+			},
+		},
+	}
+
+	for _, c := range cases {
```
t12g: Is the expectation that whenever we support a new TF kind, we should add a corresponding test case here?

test/validate_test.go
```diff
@@ -0,0 +1,277 @@
...
+	"github.com/GoogleCloudPlatform/terraform-validator/converters/google"
+	"github.com/GoogleCloudPlatform/terraform-validator/tfgcv"
+	"github.com/GoogleCloudPlatform/terraform-validator/tfplan"
+	"github.com/hashicorp/terraform/terraform"
+	"google.golang.org/api/cloudresourcemanager/v1"
+	"google.golang.org/api/option"
+)
+
+func TestValidate(t *testing.T) {
+	cfg := configure(t)
```
briantkennedy: All of this setup code should be moved after the testcase definition, it's hinders readability for it to be separated by all the testcase defs.

test/validate_test.go
```diff
@@ -0,0 +1,277 @@
...
+	baseResourceInstanceDiff := &terraform.InstanceDiff{
+		Attributes: map[string]*terraform.ResourceAttrDiff{
+			"name": &terraform.ResourceAttrDiff{New: "new-foo"},
+		},
+	}
+
+	type testcase struct {
+		kind string
+		name string
+		diff *terraform.InstanceDiff
```
briantkennedy: Is this output or input for the testcase?  Can you comment on what is getting tested against what?

test/validate_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+	"github.com/GoogleCloudPlatform/terraform-validator/converters/google"
+	"github.com/GoogleCloudPlatform/terraform-validator/tfgcv"
+	"github.com/GoogleCloudPlatform/terraform-validator/tfplan"
+	"github.com/hashicorp/terraform/terraform"
+	"google.golang.org/api/cloudresourcemanager/v1"
+	"google.golang.org/api/option"
+)
+
+func TestValidate(t *testing.T) {
```
briantkennedy: Can you add a comment on what exactly this testcase is testing?

test/validate_test.go
yukinying: I don't think table driven test mandates the need of using anonymous struct. And the struct defined here is scoped to the function. Regardless, I update that accordingly.

test/validate_test.go
yukinying: It turns out it can't be moved. The function `configure(t)` is required to be run before the test case constructions, as it read the `TEST_PROJECT`, `TEST_CREDENTIAL` and `TEST_POLICY` from environment and the project name (`TEST_PROJECT`) is required to build the test case.

test/utils.go
yukinying: Fixed. I am also thinking about just moving this to the test case so there is no need to specify it. WDYT?

test/validate_test.go
yukinying: Yes. Otherwise we should expect there are some code there is not covered by test.

morgante:
@yukinying Can you share more about the motivation for this change? I'm a little concerned about tightly coupling to the Terraform Go code because there's no commitment from Hashicorp to keep the interface stable over time (they're much more committed to CLI stability).

yukinying:
> @yukinying Can you share more about the motivation for this change? I'm a little concerned about tightly coupling to the Terraform Go code because there's no commitment from Hashicorp to keep the interface stable over time (they're much more committed to CLI stability).

The integration test currently rely on the presence of terraform CLI, which makes unit test hard to perform with better coverage. The changes are for making unit test to be run without such dependencies. 

The library itself is tightly coupled with the Terraform Go code (https://godoc.org/github.com/GoogleCloudPlatform/terraform-validator/tfplan#ComposeResources) so the test itself is making this dependency surfacing only. The problem on no commitment with Hashicorp is a valid concern thus the test can be used as a signal that the code is no longer compatible when it happens, as an update on the Go module would make the "unit" test fail. 

I also expect when the upgrade happens, these tests would fail. And then these tests will be upgraded to the new version (or dropped).

morgante:
> The integration test currently rely on the presence of terraform CLI, which makes unit test hard to perform with better coverage. The changes are for making unit test to be run without such dependencies.

I'm not sure I quite follow how it makes test coverage harder? Yes, the CLI does need to be installed but that's the only requirement.

> The library itself is tightly coupled with the Terraform Go code (https://godoc.org/github.com/GoogleCloudPlatform/terraform-validator/tfplan#ComposeResources) so the test itself is making this dependency surfacing only.

It's true that the Validator itself currently references Terraform internals, but I'd ideally like to reduce that dependency rather than expanding it. In particular, Terraform 0.12 offers a native JSON output syntax which I want to move to using so we don't have to update Validator with every internal change.

yukinying:
> I'm not sure I quite follow how it makes test coverage harder? Yes, the CLI does need to be installed but that's the only requirement.

Sorry I should have elaborated more. Currently the integration test is based on https://github.com/GoogleCloudPlatform/terraform-validator/tree/master/test/tf_templates and I believe it is not covering all kinds of TF instance (such as `google_organization_iam_policy`). By having "unit" test that has well defined structure, it can help to make tests to more precisely done and also make it easy for developers to create test cases for a particular change. This is what I referred as the coverage. 

> It's true that the Validator itself currently references Terraform internals, but I'd ideally like to reduce that dependency rather than expanding it. In particular, Terraform 0.12 offers a native JSON output syntax which I want to move to using so we don't have to update Validator with every internal change.

Would TF 0.12 be supported soon? If it is going to be there soon, say a week, I think it makes sense **NOT** to merge this as the test based on TF internal structure is a kind of work that would be thrown away.

morgante:
> Currently the integration test is based on https://github.com/GoogleCloudPlatform/terraform-validator/tree/master/test/tf_templates and I believe it is not covering all kinds of TF instance (such as google_organization_iam_policy). By having "unit" test that has well defined structure, it can help to make tests to more precisely done and also make it easy for developers to create test cases for a particular change. 

I guess I'm not convinced that directly composing the expected diff of a plan is easier than adding a bit of HCL to the Terraform config. It also means we're tightly coupled to the internal structure of how Terraform and the provider understand the resource.

> Would TF 0.12 be supported soon? If it is going to be there soon, say a week, I think it makes sense NOT to merge this as the test based on TF internal structure is a kind of work that would be thrown away.

I wouldn't expect it in the next week, but I'm hoping by end-of-month.

yukinying:
> Maybe we could instead address what complexities you see in the HCL setup?

I would like to have HCL in tests that covers all the instance resources available in https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/converters/google/mappers.go. The generation of .tf file to .state / .tfplan file can add some degree of uncertainty when there is an error. It could either be the problem of tf validator or the generation of the file.

morgante:
> I would like to have HCL in tests that covers all the instance resources available in https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/converters/google/mappers.go.

I agree that's a worthy goal and we should aim for that, but I think I'd prefer to do it with additional HCL resources.

I'm not really sure what class of errors we would see that cause breakage in the plan generation but *not* this tightly-coupled implementation. Ultimately, my main concern is the amount of maintenance burden we're taking on if we directly compose resource plans ourselves.

@chrisst Do you mind weighing in here?

yukinying:
I was reading the changes on 0.11 to 0.12. It looks like the tf plan file is in a human readable format in 0.12. In the sense, could the test be depending on plan file instead of HCL on tf file? If so, I think that would be equivalent to what I am trying to do with this pull request here and it aligns with what you suggested on not adding too much burden on test code maintenance.

morgante:
Yes, I'm still worried about potential changes to the plan format (not sure how committed HashiCorp is to that API) but I think it's still preferable to directly depending on the Go code.

Given that, I'd suggest we leave this PR on hold until the validator is upgraded to 0.12.

chrisst:
I definitely think that waiting to support 0.12 is a good move. Both state and plan output changed significantly in the upgrade to 0.12 and there is a possibility it will change again before 1.0. 
I think the only things I feel is somewhat safe from breaking interface changes are [the `helpers` modules](https://github.com/hashicorp/terraform/tree/master/helper) and [the SDK (which is grpc)](https://github.com/hashicorp/terraform/tree/master/docs/plugin-protocol).

morgante:
@yukinying Do you want to pick this back up or close it out?

yukinying:
This is no longer needed as the test has already been enhanced to use JSON form.

PR #55, opened 2019-07-23
Support Terraform 0.12 plan output for terraform-validator

### LINKED ISSUE #12: Support Terraform 0.12 plan output
See: https://github.com/hashicorp/terraform/issues/20155

kopachevsky:
Fixes #12
Terraform v0.12 validation implemented by reading plan file exported to JSON format and converting it to CAI format by creating custom FieldReader. 

Terraform validator CLI arguments does not changed, but if you pass .json file to "validate" command it will treat it as tf v0.12 version json file.

"convert" command also support TF12 as long it uses same code base as "validate"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ credentials.json
 
 # Visual Studio Code
 .vscode/
+# Intellij Golang
+.idea/
--- a/README.md
+++ b/README.md
@@ -12,6 +12,9 @@ go install .
 
 See the [Auth](#Auth) section first.
 
+
+### Steps similar both for Terraform v0.11 and v0.12 versions
+
 ```
 # The example/ directory contains a basic Terraform config for testing the validator.
 cd example/
@@ -28,9 +31,33 @@ export POLICY_PATH=/path/to/your/forseti-config-policies/repo
 # Generate a terraform plan.
 terraform plan --out=terraform.tfplan
 
+```
+
+### Terraform v0.11
+
+```
 # Validate the google resources the plan would create.
 terraform-validator validate --policy-path=${POLICY_PATH} ./terraform.tfplan
 
+# Apply the validated plan.
+terraform apply ./terraform.tfplan
+```
+
+### Terraform v0.12
+
+For 0.12 Terraform release validator required plan exported in JSON format
+
+```
+# Plan JSON representation. 
+terraform show -json ./terraform.tfplan > ./terraform.tfplan.json
+
+# Validate the google resources the plan would create.
+terraform-validator validate --tf-version 0.12 --policy-path=${POLICY_PATH} ./terraform.tfplan.json
+```
+
+### Apply validated plan
+
+```
 # Apply the validated plan.
 terraform apply ./terraform.tfplan
 ```
--- a/tfgcv/planned_assets.go
+++ b/tfgcv/planned_assets.go
@@ -17,7 +17,9 @@ package tfgcv
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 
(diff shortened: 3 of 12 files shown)


README.md
```diff
@@ -28,13 +31,37 @@ export POLICY_PATH=/path/to/your/forseti-config-policies/repo
...
+### Terraform v0.12
+
+For 0.12 Terraform release validator required plan exported in JSON format
+
+```
+# Plan JSON representation. 
+terraform show -json ./terraform.tfplan > ./terraform.tfplan.json
+
+# Validate the google resources the plan would create.
+terraform-validator validate --policy-path=${POLICY_PATH} ./terraform.tfplan.json
```
briantkennedy: I think we want to have the user specify json input / terraform version with explicit flags (eg --format=json --tf-version=0.12) since we will likely need to distinguish the version for future terraform releases.

.gitignore
```diff
@@ -12,3 +12,5 @@ credentials.json
 
 # Visual Studio Code
 .vscode/
+# Intellij Golang
+.idea
```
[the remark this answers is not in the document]
morgante: Do you mean this should be `.idea/`?

README.md
morgante: I wonder if we can auto-detect?

kopachevsky: @briantkennedy  I was also thinking about that

kopachevsky: > I wonder if we can auto-detect?
Currently it detected by file extension (.json or .tfplan) also it's not an issue to add programmatic check where input is "json" or "tfplan" format, but it will not work in case if user will pass corrupted file

kopachevsky: @briantkennedy @morgante  I think 2 params is overkill because TF12 support only json and TF11 only binary, are you ok if we'll proceed with only --tf-version=0.12?

morgante: Yes, I think proceeding with a TF version flag makes sense.

briantkennedy: In that case let's go with --tf-version and check the file extension.

kopachevsky: @briantkennedy @morgante done, please review, p.s. tf-version is required paremeter

.gitignore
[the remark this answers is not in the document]
kopachevsky: changed to .idea/

tfgcv/planned_assets.go
```diff
@@ -67,7 +51,47 @@ func ReadPlannedAssets(path, project, ancestry string) ([]google.Asset, error) {
 		return nil, errors.Wrap(err, "building google converter")
 	}
 
-	for _, r := range tfplan.ComposeResources(plan, converter.Schemas()) {
+	var resources []tfplan.Resource
+
+	switch tfVersion {
```
briantkennedy: We should error out on the default case here since we expect to be given a valid tfVersion.

README.md
```diff
@@ -28,9 +31,33 @@ export POLICY_PATH=/path/to/your/forseti-config-policies/repo
...
 terraform plan --out=terraform.tfplan
 
+```
+
+### Terraform v0.11
+
+```
 # Validate the google resources the plan would create.
-terraform-validator validate --policy-path=${POLICY_PATH} ./terraform.tfplan
+terraform-validator validate --tf-version 0.11 --policy-path=${POLICY_PATH} ./terraform.tfplan
```
briantkennedy: I think we may want to default to TF 11 but warn that the flag is not specified, then start failing in a later iteration so we don't break people.

kopachevsky: make sense

tfgcv/planned_assets.go
kopachevsky: done

kopachevsky:
@briantkennedy  applied changes to both your comments

Issue #53, opened 2019-07-18
Violation Triggered When GCS bucket_policy_only Set to "true"


FanchenBao:
**Description**
When `bucket_policy_only` is set to `true`, constraint set to __storage_bucket_policy_only.yaml__, and template set to __gcp_storage_bucket_policy_only_v1.yaml__, terraform-validator should return find NO violation. But testing at my end DOES return violation.

**Terraform Testing File: main.tf**
```
provider "google" {
  project = "some-gcp-project"
}

resource "google_storage_bucket" "sample" {
  name = "test_bucket"
  bucket_policy_only = "true"
}
```

**Possible Cause**
The CAI file converted by terraform-validator is shown below.
```
[  
  {  
    "name":"//storage.googleapis.com/test_bucket",
    "asset_type":"storage.googleapis.com/Bucket",
    "ancestry_path":"organization/../folder/../folder/../project/some-gcp-project",
    "resource":{  
      "version":"v1",
      "discovery_document_uri":"https://www.googleapis.com/discovery/v1/apis/storage/v1/rest",
      "discovery_name":"Bucket",
      "parent":"//cloudresourcemanager.googleapis.com/projects/some-gcp-project",
      "data":{  
        "lifecycle":{  
          "rule":[  

          ]
        },
        "location":"US",
        "name":"test_bucket",
        "project":"some-gcp-project",
        "storageClass":"STANDARD"
      }
    }
  }
]
```
It lacks the field "resource.data.iamConfiguration.bucketPolicyOnly.enabled", which I presume is required by the rego code in __gcp_storage_bucket_policy_only_v1.yaml__ to do the evaluation. Specifically, since the said field doesn't exist, this rego code

`bucket_policy_enabled := lib.get_default(bucket_policy_only, "enabled", null)`

leads to the `bucket_policy_enabled(bucket)` function returning `false`, which causes the rule to pass, i.e. violation is triggered.

If, in the above-mentioned code, we switch `null` to `true`, then no violation is triggered. This further suggests that the cause of the issue is most likely the disappearance of the "resource.data.iamConfiguration.bucketPolicyOnly.enabled" field.

jdyke:
Can confirm this happens on my end as well. Maybe related to https://github.com/GoogleCloudPlatform/terraform-validator/issues/52 ?

morgante:
[...]

PR #57, opened 2019-07-30
Fix/add support bucket policy only

FanchenBao:
The source file [resource_storage_bucket.go](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/third_party/terraform/resources/resource_storage_bucket.go) was updated a while back to add support for **bucket_policy_only**. Here, the corresponding files in vendor are modified to reflect such update. Now terraform-validator can support constraint `storage_bucket_policy_only.yaml`. The initial issue can be referred here #53

### DIFF
--- a/vendor/github.com/GoogleCloudPlatform/terraform-google-conversion/google/storage_bucket.go
+++ b/vendor/github.com/GoogleCloudPlatform/terraform-google-conversion/google/storage_bucket.go
@@ -183,7 +183,7 @@ func expandBucketVersioning(configured interface{}) *storage.BucketVersioning {
 	return bucketVersioning
 }
 
-func expandIamConfiguration(d *schema.ResourceData) *storage.BucketIamConfiguration {
+func expandIamConfiguration(d TerraformResourceData) *storage.BucketIamConfiguration {
 	return &storage.BucketIamConfiguration{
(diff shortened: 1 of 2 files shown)


morgante:
We need to wait for #135 to be resolved instead of directly patching vendored code.

FanchenBao:
I see. So **storage_bucket.go** in vendor depends on its upstream in [terraform-google-conversion/google/storage_bucket.go](https://github.com/GoogleCloudPlatform/terraform-google-conversion/blob/master/google/storage_bucket.go) to be fixed, which further depends on its upstream in [magic-modules/third_party/validator/storage_bucket.go](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/third_party/validator/storage_bucket.go) to be fixed.

How about **resource_storage_bucket.go** in vendor? Its upstream in [terraform-provider-google/google/resource_storage_bucket.go](https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_storage_bucket.go) has already been fixed, as well as the magic-module counterpart. Is it recommended that we fix that one first?

morgante:
We can, but it wouldn't make a difference.

FanchenBao:
Yes, you are right. I now see that everything eventually depends on magic-module, and any fixes should be directed to the hand-written code there. Thanks.

PR #70, opened 2019-08-26
Update dependencies v12

yukinying:
This would make the project use the latest v0.12 terraform package. 

It also add the use of `ConfigureBasePaths` before loading converter. It is a mandatory requirement to use converter with the latest updates.

The updates also have the following side effect:
1. Terraform v0.12 package does not have Plan.Module so the logic on parseProviderProject needs to be removed.
2. Terraform v0.12 obsoleted terraform.ReadPlan. It will always generate error. Thus the README is updated to ask reader to use the `terraform-v0.11` branch.


yukinying:
I discussed with @morgante yesterday. This needs some work to make it easy for customer to switch from v0.12 back to v0.11.

yukinying:
This pull request should be invalidated by https://github.com/GoogleCloudPlatform/terraform-validator/pull/72. I suggest keeping this pull request open for a while until #72 get merged.

PR #72, opened 2019-09-04
Update dependencies and use make prepare-v11 to switch version

yukinying:
This is based from https://github.com/GoogleCloudPlatform/terraform-validator/pull/71. 

Added the automated detection of build tag to be used in Makefile.  

This pull request would invalidate https://github.com/GoogleCloudPlatform/terraform-validator/pull/70.

Integration test would fail for TF_0_12 as the syntax of compute's diskTypes and bucket's bucketPolicyOnly has changed.

Also, ConfigureBasePaths is not added as the test coverage indicated that it is not used. It may panic if upstream changes depends on that. I would suggest that logic to be added in subsequent pull requests.

PR #73, opened 2019-09-05
Offline flag for v0.12

yukinying:
This is based from #72 to add `--offline` flag. It also makes unit test possible without connecting to GCP API. 

This pull request would invalidate #38 as #38 has no version switch logic.

### DIFF
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -86,13 +86,15 @@ type AssetResource struct {
 }
 
 // NewConverter is a factory function for Converter.
-func NewConverter(resourceManager *cloudresourcemanager.Service, project, ancestry, credentials string) (*Converter, error) {
+func NewConverter(resourceManager *cloudresourcemanager.Service, project, ancestry, credentials string, offline bool) (*Converter, error) {
 	cfg := &converter.Config{
 		Project:     project,
 		Credentials: credentials,
 	}
-	if err := cfg.LoadAndValidate(); err != nil {
-		return nil, errors.Wrap(err, "configuring")
+	if !offline {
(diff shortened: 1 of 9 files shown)


converters/google/convert.go
```diff
@@ -86,13 +86,15 @@ type AssetResource struct {
...
-func NewConverter(resourceManager *cloudresourcemanager.Service, project, ancestry, credentials string) (*Converter, error) {
+func NewConverter(resourceManager *cloudresourcemanager.Service, project, ancestry, credentials string, offline bool) (*Converter, error) {
 	cfg := &converter.Config{
 		Project:     project,
 		Credentials: credentials,
 	}
-	if err := cfg.LoadAndValidate(); err != nil {
-		return nil, errors.Wrap(err, "configuring")
+	if !offline {
+		if err := cfg.LoadAndValidate(); err != nil {
```
morgante: Is it safe to skip the loading step entirely in offline mode?

converters/google/convert.go
```diff
@@ -122,6 +124,7 @@ type Converter struct {
 
 	cfg *converter.Config
 
+	// Talk to GCP resource manager. This field would be nil in offline mode.
```
morgante: I think we should make Converter into an interface and add offline and online implementations.

morgante:
Overall looks fine, but I concur with @briantkennedy that we should change Converter to an interface with offline/online implementations.

yukinying:
I'd prefer doing that on a separate pull request (with an issue tracker). How does that sound?

morgante:
Ok, opened https://github.com/GoogleCloudPlatform/terraform-validator/issues/75 to track.

PR #76, opened 2019-09-20
allow policy and lib from a different parent folder

yukinying:
A small refactor to allow using a shared library folder and a policy folder that does not share a common parent directory.

I created `ValidateAssetsWithLibrary` as a new function because `ValidateAssets` is an exported function and I don't want to modify the function arguments as that would create a breaking change.

### DIFF
--- a/tfgcv/validate_assets.go
+++ b/tfgcv/validate_assets.go
@@ -32,11 +32,19 @@ func BuildVersion() string {
 	return buildVersion
 }
 
-// ValidateAssets instantiates GCV and audits CAI assets.
+// ValidateAssets instantiates GCV and audits CAI assets using "policies"
+// and "lib" folder under policyPath.
 func ValidateAssets(assets []google.Asset, policyPath string) (*validator.AuditResponse, error) {
+	return ValidateAssetsWithLibrary(assets,
+		filepath.Join(policyPath, "policies"),
+		filepath.Join(policyPath, "lib"))
+}
+
+// ValidateAssetsWithLibrary instantiates GCV and audits CAI assets.
+func ValidateAssetsWithLibrary(assets []google.Asset, policyPath, policyLibraryDir string) (*validator.AuditResponse, error) {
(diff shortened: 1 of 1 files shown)


tfgcv/validate_assets.go
```diff
@@ -32,11 +32,19 @@ func BuildVersion() string {
...
+// ValidateAssets instantiates GCV and audits CAI assets using "policies"
+// and "lib" folder under policyPath.
 func ValidateAssets(assets []google.Asset, policyPath string) (*validator.AuditResponse, error) {
+	return ValidateAssetsWithLibrary(assets,
+		filepath.Join(policyPath, "policies"),
+		filepath.Join(policyPath, "lib"))
+}
+
+// ValidateAssetsWithLibrary instantiates GCV and audits CAI assets.
+func ValidateAssetsWithLibrary(assets []google.Asset, policyPath, policyLibraryDir string) (*validator.AuditResponse, error) {
```
morgante: Do we need to export this?

tfgcv/validate_assets.go
```diff
@@ -32,11 +32,19 @@ func BuildVersion() string {
 	return buildVersion
 }
 
-// ValidateAssets instantiates GCV and audits CAI assets.
+// ValidateAssets instantiates GCV and audits CAI assets using "policies"
+// and "lib" folder under policyPath.
 func ValidateAssets(assets []google.Asset, policyPath string) (*validator.AuditResponse, error) {
```
morgante: Could we give the two policyPath arguments different names since they represent different values now? Maybe change this policyPath to `policyRootPath`?

tfgcv/validate_assets.go
yukinying: Yes. I have a project that need to call this function directly.

tfgcv/validate_assets.go
yukinying: Done.

PR #82, opened 2019-10-03
extract ancestry discovery logic into an interface

### LINKED ISSUE #75: Make offline mode a separate type
https://github.com/GoogleCloudPlatform/terraform-validator/pull/73#pullrequestreview-288819696

yukinying:
This fixes #75 with a small change of scope. Instead of making converter an interface, this creates a new interface called AncestryManager which has a GetAncestry method. 

The implementation for online mode uses resource manager and a cache. This works the same way as when --offline=false. 

Test has been added with a mock of resource manager API server for GetAncestry call.

Also, the need of calling `converter.LoadAndValidate()` is removed as the upstream dependency (`terraform_google_conversion`) do not requires the clients to be constructed for the TF module supported. Please review that as well.

I would also like to get help to see if the names for the interface and methods make sense. I cannot make up a better name.

### DIFF
--- a/ancestrymanager/ancestrymanager.go
+++ b/ancestrymanager/ancestrymanager.go
@@ -17,7 +17,7 @@ type AncestryManager interface {
 	GetAncestry(project string) (string, error)
 }
 
-type ancestryManager struct {
+type onlineAncestryManager struct {
 	// Talk to GCP resource manager. This field would be nil in offline mode.
 	resourceManager *cloudresourcemanager.Service
 	// Cache to prevent multiple network calls for looking up the same project's ancestry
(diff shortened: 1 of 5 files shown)


yukinying:
Another reason I want to abstract the ancestry discovery logic as an interface is that there are more than one way to find ancestry. E.g. if you have access to CAI, you could probably find the ancestry data there (correct me if I am wrong). In that sense, I would not need to enable resource manager API.

morgante:
This definitely looks like a change in a good direction!

I think I'd still prefer that offline and online mode be represented by two different types/structs (which both implement the ancestryManager interface). Otherwise every access to resource manager will need a nil check (or potentially introduce runtime errors).

yukinying:
> I think I'd still prefer that offline and online mode be represented by two different types/structs (which both implement the ancestryManager interface). Otherwise every access to resource manager will need a nil check (or potentially introduce runtime errors).

I have thought about it for a while. In any case we need to perform nil check before calling `am.resourceManager.Projects.GetAncestry(...)`, so offline and online ancestry manager implementation is pretty much the same.

morgante:
> I have thought about it for a while. In any case we need to perform nil check before calling am.resourceManager.Projects.GetAncestry(...), so offline and online ancestry manager implementation is pretty much the same.

Why do we need the nil check in both cases?

In offline mode, we should simply never be making that call.
In online mode, we need a resource manager to initialize the ancestryManager.

yukinying:
True. I have made the changes.

ancestrymanager/ancestrymanager.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// New returns AncestryManager that can be used to fetch ancestry information for a project.
+func New(ctx context.Context, project, ancestry string, offline bool, opts ...option.ClientOption) (AncestryManager, error) {
+	if offline {
+		return &offlineAncestryManager{project: project, ancestry: ancestry}, nil
+	}
+	am := &onlineAncestryManager{ancestryCache: map[string]string{}}
+	am.store(project, ancestry)
+	if !offline {
```
morgante: We don't need this conditional block.

yukinying: fixed.

Issue #52, opened 2019-07-12
Add support for GCS IAM policies


morgante:
Resources:
- google_storage_bucket_iam_binding
- google_storage_bucket_iam_policy

PR #84, opened 2019-10-23
Added mappers for storage bucket iam

onetwopunch:
Fixes: https://github.com/GoogleCloudPlatform/terraform-validator/issues/52

### DIFF
--- a/converters/google/mappers.go
+++ b/converters/google/mappers.go
@@ -116,6 +116,24 @@ func mappers() map[string][]mapper {
 				merge:   converter.MergeProjectIamMember,
 			},
 		},
+		"google_storage_bucket_iam_policy": {
+			{
+				convert: converter.GetBucketIamPolicyCaiObject,
+				merge:   converter.MergeBucketIamPolicy,
+			},
+		},
+		"google_storage_bucket_iam_binding": {
+			{
+				convert: converter.GetBucketIamBindingCaiObject,
+				merge:   converter.MergeBucketIamBinding,
+			},
+		},
+		"google_storage_bucket_iam_member": {
+			{
+				convert: converter.GetBucketIamMemberCaiObject,
+				merge:   converter.MergeBucketIamMember,
+			},
+		},
 	}
 }
 
--- a/test/convert_test.go
+++ b/test/convert_test.go
@@ -39,6 +39,7 @@ var conversionTests = []struct {
 	{"firewall", "compute.googleapis.com/Firewall"},
 	{"instance", "compute.googleapis.com/Instance"},
 	{"bucket", "storage.googleapis.com/Bucket"},
+	{"bucket_iam", "storage.googleapis.com/Bucket:IAM"},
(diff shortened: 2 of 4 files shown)


onetwopunch:
@morgante can you review this when you have a moment?

converters/google/mappers.go
```diff
@@ -116,6 +116,24 @@ func mappers() map[string][]mapper {
 				merge:   converter.MergeProjectIamMember,
 			},
 		},
+		"google_storage_bucket_iam_policy": {
+			{
+				convert: converter.GetBucketIamPolicyCaiObject,
+				merge:   converter.MergeBucketIamPolicy,
```
morgante: Doesn't look like this is available in the vendored code? Do we need to update conversion?

onetwopunch: It's rebased on master. This function is [here](https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/vendor/github.com/GoogleCloudPlatform/terraform-google-conversion/google/storage_bucket_iam.go#L17)

test/convert_test.go
```diff
@@ -63,13 +64,19 @@ func TestConvert(t *testing.T) {
...
 
-		assetsByType := make(map[string][]google.Asset)
+		assetsByType := make(map[string]google.Asset)
 		for _, a := range assets {
-			assetsByType[a.Type] = append(assetsByType[a.Type], a)
+			t := a.Type
+			// NOTE: Since IAM policy assets will have the same type, we'll add a
+			// suffix to map it to the right JSON fixture
+			if a.IAMPolicy != nil {
+				fmt.Println("POLICY")
```
morgante: This should probably either be dropped or use t.Log/Logf.

test/convert_test.go
```diff
@@ -63,13 +64,19 @@ func TestConvert(t *testing.T) {
 			t.Fatalf("unmarshaling: %v", err)
 		}
 
-		assetsByType := make(map[string][]google.Asset)
+		assetsByType := make(map[string]google.Asset)
 		for _, a := range assets {
-			assetsByType[a.Type] = append(assetsByType[a.Type], a)
+			t := a.Type
```
morgante: Let's not override t.

```suggestion
			aType := a.Type
```

test/convert_test.go
onetwopunch: Good catch

test/convert_test.go
onetwopunch: Oh good call haha...totally forgot that's a special var.

onetwopunch:
@morgante fixed a typo and rebased after you approved. Would you mind merging when you have a moment?

yukinying:
@onetwopunch Would you also consider adding the new module in the README.md?

onetwopunch:
@yukinying Sure :) https://github.com/GoogleCloudPlatform/terraform-validator/pull/92

Issue #86, opened 2019-11-02
folder does not get expanded in google_folder_iam_member


yukinying:
Step to reproduce: 

* Create a simple Terraform configuration like the one in documentation.
```
resource "google_folder" "department1" {
  display_name = "Department 1"
  parent       = "organizations/1234567"
}

resource "google_folder_iam_member" "admin" {
  folder  = "${google_folder.department1.name}"
  role    = "roles/editor"
  member  = "user:alice@gmail.com"
}
```

* run terraform plan, terraform show.
* add a debug statement around https://github.com/GoogleCloudPlatform/terraform-validator/blob/1fe01ae2fc848730682acd7c4cab7b5cb1bbc970/cmd/validate.go#L52 to print the assets
* notice that the name of the assets is in the form of `//cloudresourcemanager.googleapis.com/placeholder-qbilI01z"`

yukinying:
Since https://github.com/GoogleCloudPlatform/terraform-google-conversion/blob/6c802a86e5c435b98efe6e08d03aaff2c680e906/google/folder_iam.go#L41 should expand the {{folder}} field, this seems like a problem on parsing {{folder}}.

I have also tested that if I replace `"${google_folder.department1.name}"` with a static value, then the expansion logic works correctly.

yukinying:
// cc @morgante

morgante:
Unfortunately there isn't any way for us to know the actual folder ID from the plan file, since the folder ID isn't specified until it is created.

melinath:
Based on Morgante's comment it seems like this isn't going to be feasible - and it's not entirely clear to me what problem this was causing. If anyone is having problems due to this behavior, please open an issue with details!

Issue #81, opened 2019-10-03
panic: runtime error: index out of range [0] with length 0


jenshonkan84:
Hi

I get this error when running the validator against a plan file that has a google_storage_bucket resource in it.
```
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/GoogleCloudPlatform/terraform-google-conversion/google.GetStorageBucketApiObject(0x285e000, 0xc000121d20, 0xc0000fcf00, 0x2470baf, 0x21, 0xc000c8d320)
        /home/jens_hornstrom_seb_se/terraform-validator/vendor/github.com/GoogleCloudPlatform/terraform-google-conversion/google/storage_bucket.go:80 +0x10e4
github.com/GoogleCloudPlatform/terraform-google-conversion/google.GetStorageBucketCaiObject(0x285e000, 0xc000121d20, 0xc0000fcf00, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /home/jens_hornstrom_seb_se/terraform-validator/vendor/github.com/GoogleCloudPlatform/terraform-google-conversion/google/storage_bucket.go:25 +0xdc
github.com/GoogleCloudPlatform/terraform-validator/converters/google.(*Converter).AddResource(0xc000393bf0, 0x2854e80, 0xc00025d980, 0x0, 0x0)
        /home/jens_hornstrom_seb_se/terraform-validator/converters/google/convert.go:157 +0x14c
github.com/GoogleCloudPlatform/terraform-validator/tfgcv.ReadPlannedAssets(0x7fff8c0e872e, 0x16, 0x0, 0x0, 0x0, 0x0, 0xc000229c00, 0x0, 0x0, 0x0, ...)
        /home/jens_hornstrom_seb_se/terraform-validator/tfgcv/planned_assets.go:101 +0x516
github.com/GoogleCloudPlatform/terraform-validator/cmd.glob..func6(0x36f1ce0, 0xc000329520, 0x1, 0x2, 0x0, 0x0)
        /home/jens_hornstrom_seb_se/terraform-validator/cmd/validate.go:52 +0x9b
github.com/spf13/cobra.(*Command).execute(0x36f1ce0, 0xc0003294e0, 0x2, 0x2, 0x36f1ce0, 0xc0003294e0)
        /home/jens_hornstrom_seb_se/terraform-validator/vendor/github.com/spf13/cobra/command.go:762 +0x460
[...]

PR #88, opened 2019-11-04
add unit tests for ReadPlannedAssets based on examples from Terraform Google Provider

yukinying:
The purpose of the the test is to uncover potential issues on the supported Terraform configuration, and to allow adding new configuration comes with coverage. This can be also used as a way to create regression tests on runtime error bugs like https://github.com/GoogleCloudPlatform/terraform-validator/issues/81 or feature request validation like https://github.com/GoogleCloudPlatform/terraform-validator/issues/53.  

In this pull requests, 12 test cases are constructed using examples from Terraform Google Provider:

```
google_compute_disk
google_compute_firewall
google_compute_instance
google_organization_iam_binding
google_organization_iam_member
google_organization_iam_policy
google_project
google_project_iam_binding
google_project_iam_member
google_project_iam_policy
google_sql_database_instance
google_storage_bucket
```

I haven't added the google_folder_* configuration as test cases because the folder expansion logic is not working (https://github.com/GoogleCloudPlatform/terraform-validator/issues/86).

The tests also unveiled that the disk image path resolving logic is not working correctly unless the image is provided in the form of `projects/<project>/global/images/<image>` or `global/images/<image>`. I rewrite the example configuration to use this image format in `google_compute_disk` and `google_compute_instance`.

I have also written a few more test cases with more comprehensive coverage. Those may need more time from reviewers to validate if the resultant assets is in the expected form, thus they will be provided in other pull requests after this one get through.


morgante:
Please make sure `make test` passes. Otherwise LGTM, we can converge in followup (#91).

Issue #91, opened 2019-11-05
Unify unit and integration tests


morgante:
#88 introduced tests which are very similar to existing [integration tests](https://github.com/GoogleCloudPlatform/terraform-validator/tree/master/test).

Instead of maintaining these separately, we should:

1. Make a unified test/ folder which combines this with the existing test data
2. Provide two modes:
    - In unit test mode, it uses pre-generated plan JSON to validate the CAI conversion (like what you have in this PR)
    - In integration test mode, it uses the provided HCL to generate the plan, then proceeds to the unit tests from (1)

yukinying:
This is now fixed.

PR #94, opened 2019-11-08
refactor integration test

yukinying:
This is for preparation of merging unit tests with integration tests as discussed in https://github.com/GoogleCloudPlatform/terraform-validator/issues/91. 

The changes includes
1. moving /test/{*_templates,sample_policies} to /testdata. This allow unit test code to refer to the testdata path more intuitively.
2. refactor integration test logic into one single file: /test/cli_test.go
3. make integration tests run in parallel (it takes 6s to complete all tests now).
4. make each integration test runs with generated file in a generated tmp directory. This allow separating resources into individual integration test.
5. update /testdata/templates/*.tf to include the provider configuration, so main.tf is not needed.
6. merged data and cfg structure as they are both depending on environment variable. The structure is moved to a global struct (similar to const, but depends on env var) and initialized at test start time.
7. provide a "fakehome" directory as terraform touches it.
8. improved the table driven test structure to allow future addition of validation rules per resource.

Caveats
1. I make each command be run in a separate function to make the code with better readibility. This may introduce repeated code. I am also running out of idea on how to give better names to functions so pleaes pardon the bad naming, or suggest something better.


yukinying:
Can I get the error trace from the build? I have tested this manually with `make test` and `make test-`integration` and both of them has reported no issue locally.

yukinying:
I got access to the build error. It looks like we are building with Go 1.11. The latest version of Go support String() in exec.Cmd. 

Do we have a plan to upgrade https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/Dockerfile#L1 to use Go 1.13?

In the meanwhile, I will change the tests to make it works with Go 1.11.

yukinying:
The integration test is now passing.

morgante:
> Do we have a plan to upgrade https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/Dockerfile#L1 to use Go 1.13?

Happy to look at a PR.

test/cli_test.go
```diff
@@ -0,0 +1,379 @@
...
+		Ancestry: ancestry,
+	}
+}
+
+// TestCLI tests the "convert" and "validate" subcommand against a generated .tfplan file.
+func TestCLI(t *testing.T) {
+	// Define the reusable rules to be use for the test cases.
+	type rule struct {
+		name            string
+		wantError       bool
```
morgante: We should distinguish violations and errors.

```suggestion
		wantViolation       bool
```

test/cli_test.go
```diff
@@ -0,0 +1,379 @@
...
+			"BillingAccountName": "012345-567890-ABCDEF",
+		},
+		Ancestry: ancestry,
+	}
+}
+
+// TestCLI tests the "convert" and "validate" subcommand against a generated .tfplan file.
+func TestCLI(t *testing.T) {
+	// Define the reusable rules to be use for the test cases.
+	type rule struct {
```
morgante: Let's keep naming consistent:

```suggestion
	type constraint struct {
```

test/cli_test.go
```diff
@@ -0,0 +1,379 @@
...
+		terraformPlan(t, "terraform", dir, name+".tfplan")
+		payload := terraformShow(t, "terraform", dir, name+".tfplan")
+		saveFile(t, dir, name+".tfplan.json", payload)
+	}
+}
+
+func terraformInit(t *testing.T, executable, dir string) {
+	wantError := false
+	cmd := exec.Command(executable, "init", "-input=false", dir)
+	cmd.Env = []string{"HOME=" + filepath.Join(dir, "fakehome")}
```
morgante: There seems to be a lot of repeated code between `terraformInit`, `terraformPlan`, etc.

Could we consolidate with a `terraformExec` function and simply call that multiple times with different commands?

test/cli_test.go
```diff
@@ -0,0 +1,379 @@
...
+func tfvConvert(t *testing.T, dir, tfplan string, offline bool) []byte {
+	executable := tfvBinary
+	wantError := false
+	args := []string{"convert", "--project", data.Provider["project"]}
+	if offline {
+		args = append(args, "--ancestry", data.Ancestry)
+	}
+	args = append(args, tfplan)
+	cmd := exec.Command(executable, args...)
+	cmd.Env = []string{"GOOGLE_APPLICATION_CREDENTIALS=" + data.Provider["credentials"]}
```
morgante: If `TEST_CREDENTIALS` isn't explicitly set, then we're already relying on application default credentials and shouldn't override.

So let's make this line conditional on `data.Provider["credentials"]` not being nil.

testdata/templates/sql.tf
```diff
@@ -1,3 +1,8 @@
+provider "google" {
+  version     = "~> {{.Provider.version}}"
+  credentials = "{{.Provider.credentials}}"
```
morgante: I think this prevents the usage of Application Default Credentials?

Could we make the injection of `credentials = ` conditional on an explicit `TEST_CREDENTIALS` file being provided?

test/convert_test.go
```diff
@@ -1,147 +0,0 @@
...
-
-const samplePolicyPath = "sample_policies"
-
-var conversionTests = []struct {
-	name      string
-	assetType string
-}{
-	{"disk", "compute.googleapis.com/Disk"},
-	{"project", "cloudresourcemanager.googleapis.com/Project"},
-	{"project_billing_info", "cloudbilling.googleapis.com/ProjectBillingInfo"},
```
morgante: Looks like this test got removed. More generally, how will we deal with cases (like this) where one resource (the project) actually generates two assets (project and project billing info)? Let's make sure we don't lose this test in the conversion.

test/convert_test.go
```diff
@@ -1,147 +0,0 @@
...
-	{"firewall", "compute.googleapis.com/Firewall"},
-	{"instance", "compute.googleapis.com/Instance"},
-	{"bucket", "storage.googleapis.com/Bucket"},
-	{"bucket_iam", "storage.googleapis.com/Bucket:IAM"},
```
morgante: Did we remove the bucket IAM test? Let's be sure to keep it, particularly as I want to see how you plan to handle IAM tests in the new approach.

testdata/templates/project.json
```diff
@@ -0,0 +1,40 @@
+[
+  {
+    "name": "//cloudbilling.googleapis.com/projects/{{.Provider.project}}/billingInfo",
+    "asset_type": "cloudbilling.googleapis.com/ProjectBillingInfo",
```
yukinying: Note: billing is covered here.

yukinying: and testdata/templates/project_billing_info.json need to be removed.

test/convert_test.go
yukinying: Yes it needs to be added back. Sorry for the oversight.

test/convert_test.go
yukinying: Billing is combined with the project test as it is handled via one single TF file.

testdata/templates/project.json
yukinying: removed.

test/convert_test.go
yukinying: Test case for bucket_iam is added.

testdata/templates/sql.tf
yukinying: added.

test/cli_test.go
yukinying: added.

test/cli_test.go
yukinying: yes. Fixed.

test/cli_test.go
yukinying: yes. Fixed.

test/cli_test.go
yukinying: yes. Fixed.

PR #95, opened 2019-11-08
add unit tests to add more coverages

yukinying:
Adding more coverage on fields listed in the terraform google provider on the following kinds of resource:
* compute_firewall
* compute_instance
* sql_database_instance
* storage_bucket

The TF files are generated manually using values in the form of "test-<fieldname>" when possible. Some of those are using the values from examples. Some are using valid values if Terraform Google provider check for validity of such.

It is expected these tests could catch the case of panic if a field is not handled properly.


tfgcv/testdata/full_compute_firewall.tf
```diff
@@ -0,0 +1,51 @@
...
+
+  allow {
+    protocol = "icmp"
+  }
+  allow {
+    protocol = "tcp"
+    ports    = ["80", "8080", "1000-2000"]
+  }
+  description        = "test-description"
+  destination_ranges = ["test-destination_ranges1", "test-destination_ranges2"]
```
morgante: I'm not sure about whether we should be using these hard-coded `test-*` values? In the future, there might be provider or conversion checks which ensure you're putting in valid values, and I'm not sure we actually benefit.

It seems like we'd be better off using an actual (but known) CIDR block here (or known zones, etc.).

yukinying: Agree. This is an initial attempt to build a sketelon of stuff that should be tested. I currently rely on `terraform plan` to complain if a value need to be changed. I also refer to the example in the documentation occasionally. 

I can fix the CIDR and zones.

morgante: Got it, I can call out. the ones I think need changing

tfgcv/testdata/full_compute_firewall.tf
```diff
@@ -0,0 +1,51 @@
...
+  network = "${google_compute_network.default.name}"
+
+  deny {
+    protocol = "icmp"
+  }
+  deny {
+    protocol = "tcp"
+    ports    = ["80", "8080", "1000-2000"]
+  }
+  source_ranges           = ["test-source_range1", "test-source_range2"]
```
morgante: Here.

tfgcv/testdata/full_compute_instance.tf
```diff
@@ -0,0 +1,121 @@
...
+    # Got: "boot_disk.0.kms_key_self_link": conflicts with boot_disk.0.disk_encryption_key_raw
+    # kms_key_self_link = "test-kms_key_self_link"
+    # TODO: report a bug.
+    # Got: An argument named "mode" is not expected here.
+    # mode = "READ_ONLY"
+    source = "test-source"
+  }
+  network_interface {
+    access_config {
+      nat_ip = "test-nat_ip"
```
morgante: I think this should be a real IP? Or, even better, omit it entirely (as I think that's the best scenario to test/verify works correctly).

tfgcv/testdata/full_compute_instance.tf
```diff
@@ -0,0 +1,121 @@
...
+    }
+    access_config {
```
morgante: I don't know if multiple access_config_blocks are actually allowed?

yukinying:
> Could we try starting from a Terraform config which can be successfully applied (even in an isolated test project/non-reproducably)?

That would be the ones with example_*. Also we can change these tests to make it works in a real projecet.

But it is also important to make sure all fields are presented (or presented when all tests are combined). I can see that the issues which caused the panics are mostly related to fields that are provided with wrong types in terraform-google-conversion. Empty value would make the issues subsided.

tfgcv/testdata/full_compute_instance.tf
yukinying: yes they are allowed... at least magic module does not block them in terraform google provider.

morgante: Ok, but I'm not sure they are really allowed by the GCP API.

yukinying: FYI. https://www.terraform.io/docs/providers/google/r/compute_instance.html

```
access_config - (Optional) Access configurations, i.e. IPs via which this instance can be accessed via the Internet. Omit to ensure that the instance is not accessible from the Internet. If omitted, ssh provisioners will not work unless Terraform can send traffic to the instance's network (e.g. via tunnel or because it is running on another cloud instance on that network). This block can be **repeated multiple times**. Structure documented below.
```

morgante:
@yukinying Can you resolve conflicts?

yukinying:
Code updated with the following changes:

1. moved TestReadPlannedAssetsCoverage from tfgcv/planned_assets_test.go to test/read_test.go. As the refactoring of integration test has been merged, this pull request is already updated with those changes. In particular, the template generation logic is in test so the test needs to be moved to there.

2. the test case in TestCLI would run through both offline and online mode without having the test case specifying it. 

3. the test case in TestCLI would use `[]constraint{alwaysViolate}` by default.

4. unit test has been updated to use template.

yukinying:
Caveat: as there are more tests running, the integration test now takes 22s in my local machine instead of 6s.

yukinying:
> Can you resolve conflicts?

Yes it should have resolved.

tfgcv/testdata/full_compute_instance.tf
yukinying: updated with valid IP.

tfgcv/testdata/full_compute_firewall.tf
yukinying: updated with valid IP ranges.

tfgcv/testdata/full_compute_firewall.tf
yukinying: updated with valid IP ranges.

yukinying: zone, region and location field are updated to use valid values.

morgante:
The code looks ok, but it seems tests are now taking a very long time. CI has been running for over 10 minutes: https://pantheon.corp.google.com/cloud-build/builds/6c4dc6ed-f084-4424-9afc-f3bf0cd59440;step=0?project=cloud-foundation-cicd

yukinying:
Could we use a faster machine for integration test? I also suspect there may be multiple build running and this caused the slowness.

morgante:
> Could we use a faster machine for integration test?

Yes, we could configure that in the Cloud Build yaml.

PR #96, opened 2019-11-09
support google_container_cluster and google_container_node_pool

yukinying:


tfgcv/testdata/full_container_cluster.tf
```diff
@@ -0,0 +1,133 @@
...
+  # ip_allocation_policy.subnetwork_name
+  # region
+  # zone
+
+  # TODO: problematic fields
+  # node_pool
+
+  name = "test-cluster"
+
+  additional_zones = ["test-additional_zones"]
```
morgante: Not sure if we should be putting invalid values in here. At the very least, should probably be real zones.

tfgcv/testdata/full_container_node_pool.tf
```diff
@@ -0,0 +1,61 @@
...
+  # node_config.workload_metadata_config
+  # node_locations
+
+  autoscaling {
+    min_node_count = 42
+    max_node_count = 1337
+  }
+  cluster = "test-cluster"
+  initial_node_count = 42
+  location = "test-location"
```
morgante: We should probably stick to real locations, even in tests.

PR #98, opened 2019-11-16
use -mod=vendor in build

yukinying:
The build is not using vendor directory and that make every build spent more than 2 minutes just on downloading dependencies in Cloud Build. 

I also suspect that the -mod flag needs to be provided before other flags.


yukinying:
This build finished in 9 min, and saved around 1.5 min in downloading dependencies.

PR #99, opened 2019-11-16
use N1_HIGHCPU_8 in Cloud Build

yukinying:


yukinying:
This build finished in 3 min 5 sec, which is nearly a 200% improvement.

PR #100, opened 2019-11-16
use cft/developer-tools:0.5.0 in Cloud Build

yukinying:
It looks like Cloud Foundation Toolkit (CFT) is built using the same image with different version. Interestingly the build in CFT can pull the images in less than a minute. Upgrading the version used here to 0.5.0 may speed up the build.

Reference: 
https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/98a5af4e94561f062e0ceb0e62a5d3f990ff35fd/infra/cloudbuild/test.devtools.yaml#L16


yukinying:
This build does not have any improvement yet.

morgante:
I suspect the download speed might just be an artifact of caching (based on which build had recently been run on other CFT repos).

PR #102, opened 2019-11-19
Short test

yukinying:
This make unit test runs correctly via `make test`. Previously it was skipped in unit test stage and get run again in integration test stage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,14 +13,14 @@ prepare-v11:
 	go mod vendor
 
 test:
-	# Skip integration tests in ./test/
-	GO111MODULE=on go test -mod=vendor -tags=$(BUILDTAGS) `go list -mod=vendor -tags=$(BUILDTAGS) ./... | grep -v terraform-validator/test`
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -13,14 +13,14 @@ prepare-v11:
 	go mod vendor
 
 test:
-	# Skip integration tests in ./test/
-	GO111MODULE=on go test -mod=vendor -tags=$(BUILDTAGS) `go list -mod=vendor -tags=$(BUILDTAGS) ./... | grep -v terraform-validator/test`
+	# Skip integration tests in ./test/ using -short flag
+	GO111MODULE=on go test -short -mod=vendor -tags=$(BUILDTAGS) ./...
```
morgante: Coes `-mod=vendor` need to be the first argument still?

yukinying: Now I have more data in the build pipeline, I think this is not true and is probably misobservation from myself with some other changes that I have missed.

Issue #110, opened 2019-11-26
Child_modules are nested, and only the first level is parsed


yukinying:
Per https://www.terraform.io/docs/internals/json-format.html#plan-representation
```
    "child_modules": [
      // Each entry in "child_modules" has the same structure as the root_module
      // object, with the additional "address" property shown below.
      {
        // [Other fields omitted] 
       
        // Each module object can optionally have its own
        // nested "child_modules", recursively describing the
        // full module tree.
        "child_modules": [ ... ],
      }
```

However, https://github.com/GoogleCloudPlatform/terraform-validator/blob/0c219acc485d54b2a62537bdd4072d678fbe33d8/tfplan/json_plan.go#L27-L36 indicated that only the first level of child modules are parsed. This seems to be problematic as nested child modules are very common in Terraform configurations.

yukinying:
And it looks like the child module parsing is done in https://github.com/GoogleCloudPlatform/terraform-validator/blob/0c219acc485d54b2a62537bdd4072d678fbe33d8/tfplan/json_plan.go#L182-L188

morgante:
Good catch, thank you! This is definitely a big issue and we should probably add some tests to ensure it's addressed.

PR #111, opened 2019-11-26
support nested child modules

yukinying:
This fixes https://github.com/GoogleCloudPlatform/terraform-validator/issues/110

### DIFF
--- a/tfplan/json_plan.go
+++ b/tfplan/json_plan.go
@@ -23,28 +23,33 @@ import (
 	"github.com/pkg/errors"
 )
 
+const maxChildModuleLevel = 100
+
 // jsonPlan structure used to parse Terraform 12 plan exported in json format by 'terraform show -json ./binary_plan.tfplan' command.
 type jsonPlan struct {
 	PlannedValues struct {
 		RootModules struct {
-			Resources    []jsonResource
-			ChildModules []struct {
-				Address   string
-				Resources []jsonResource
-			} `json:"child_modules"`
+			Resources    []jsonResource `json:"resources"`
+			ChildModules []childModule  `json:"child_modules"`
 		} `json:"root_module"`
 	} `json:"planned_values"`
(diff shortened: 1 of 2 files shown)


tfplan/json_plan_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestReadJSONResources(t *testing.T) {
+	data := []byte(`
+{
+  "planned_values": {
+    "root_module": {
+      "child_modules": [
```
morgante: Could we add a deeply nested test (2+ levels) just to confirm we don't introduce subtle bugs with recursion?

tfplan/json_plan.go
```diff
@@ -167,25 +171,36 @@ func attributes(value interface{}, address []string, res map[string]string, sche
...
-			result = append(result, resource)
-		}
+		result = append(result, resourcesFromModule(&module, 0)...)
 	}
-
 	return result, nil
 }
+
+func resourcesFromModule(module *childModule, level int) []jsonResource {
+	if level > maxChildModuleLevel {
```
morgante: Is there a practical reason we need a max?

yukinying: This is to prevent potential denial of service from user input, which is becoming pretty common kinds of vulnerability.

tfplan/json_plan_test.go
yukinying: done.

yukinying: actually the test is on 2 level already (level 1 has no resource, level 2 has one), now I added an extra level so as to make it more resilient.

tfplan/json_plan.go
yukinying: I relaxed the max to 10000 and add a log message if this condition is hit.

PR #114, opened 2019-12-04
Improve handling for google_project resource ancestry.

### LINKED ISSUE #112: Issues with deploying google_project resource
We are encountering an issue with deploying the `google_project` resource in our pipelines.  The error we encounter is as follows:

`Error: converting tfplan to CAI assets: converting resource google_project: augmenting asset: project: required field is not set`

I understand that generally there would be nothing this resource could violate by itself however for our CI/CD pipelines there will be times when users will simply be deploying just a single project (or folder).  Is there currently a way for us to bypass the project field when the TFPlan only contains the `google_project` or `google_folder` resource?  

I apologize ahead of time if this has already been addressed I just was unable to find anything in my searches.

Google Project TF: https://www.terraform.io/docs/providers/google/r/google_project.html
Google Folder TF: https://www.terraform.io/docs/providers/google/r/google_folder.html

morgante:
Currently, we don't try to intelligently build the resource ancestry for projects. This improves on that:

1. Add an example `google_project` config.
2. Fix #112 by retrieving the project_id directly from the resource (if possible)
3. Attempt to build the ancestry for projects without relying on the GetAncestry API
4. Also updated README to focus on Terraform 0.12.
5. Add new test cases for projects with explicit `org_id` and `folder_id`.


morgante:
It's not clear why tests are suddenly failing, will investigate.

ancestrymanager/ancestrymanager.go
```diff
@@ -4,20 +4,120 @@ package ancestrymanager
...
+	tfData      converter.TerraformResourceData
+	hasAsset    bool
+	asset       converter.Asset
+}
+
+// GetAncestryOption sets options for an ancestry tequest
+type GetAncestryOption func(*getAncestryRequest)
+
+// TerraformResource sets the Terraform resource for a getAncestry request
+func TerraformResource(tfData converter.TerraformResourceData) GetAncestryOption {
```
yukinying: It would be easier for reader to figure this out if the function can be named in the form of `With...` or `For...`, such as `ForTerraformResource`.

converters/google/convert.go
```diff
@@ -192,11 +192,11 @@ func (c *Converter) Assets() []Asset {
 
 // augmentAsset adds data to an asset that is not set by the conversion library.
 func (c *Converter) augmentAsset(tfData converter.TerraformResourceData, cfg *converter.Config, cai converter.Asset) (Asset, error) {
-	project, err := getProject(tfData, cfg)
+	project, err := getProject(tfData, cfg, cai)
 	if err != nil {
 		return Asset{}, err
 	}
-	ancestry, err := c.ancestryManager.GetAncestry(project)
+	ancestry, err := c.ancestryManager.GetAncestry(project, ancestrymanager.TerraformResource(tfData), ancestrymanager.Asset(cai))
```
yukinying: Could this implemented as a new function (e.g. `GetAncestryFromResource`) so there is no need to pass in variables of Option type?

morgante: Yeah, looking at this now I think that could be a better approach. Will update.

morgante:
@yukinying Please take another look

go.mod
```diff
@@ -2,17 +2,21 @@ module github.com/GoogleCloudPlatform/terraform-validator
 
 require (
 	github.com/GoogleCloudPlatform/terraform-google-conversion v0.0.0-20191115234250-56efb5b73dd1
+	github.com/coreos/go-systemd v0.0.0-20181012123002-c6f51f82210d // indirect
```
yukinying: Are these dependencies getting into vendor directory?

PR #116, opened 2019-12-06
Fix error with glog not parsing flags early enough

morgante:


yukinying:
Interesting. Is this an attempt to make certain flag usable inside init function? I haven't dig too deep into this but it looks like there has been a lot of discussion on go community that flag should not be parsed inside init function.

morgante:
@yukinying This is to prevent an issue with glog raising errors due to logs we output quite early (including in the test init).

Note that this is a change to the *test* init, not the main code paths. It's only an issue for tests.

PR #128, opened 2020-02-26
Add support for google_bigquery_dataset.

xingao267:
Fixes #127 #23

PR #129, opened 2020-02-26
Update terraform-google-conversion to latest.

xingao267:
- Updated terraform-google-conversion to latest version.
- Updated terraform-google-provider to latest version.
- Updated Terraform Google Provider version to 3.0.0 in tests.
- Updated google.golang.org/api to v0.16.0 to be consistent with the one used by terraform-google-provider
- Fixed code and tests due to updating the above dependencies.


xingao267:
@morgante @danawillow Bring up the `terraform-google-conversion` library to latest first so if there is fix in `terraform-google-conversion` needed for https://github.com/GoogleCloudPlatform/terraform-validator/pull/128. I can use the fix in https://github.com/GoogleCloudPlatform/terraform-validator/pull/128 easily.

Need some help regarding the failing tests. The reason might be related to the same cause similar in https://github.com/GoogleCloudPlatform/terraform-validator/pull/128

xingao267:
@morgante Yea I fixed the tests that I could but still have few more left which I might need some help from you.

morgante:
I suspect the other error you're seeing is related to the diff between these two files: [Conversion](https://github.com/GoogleCloudPlatform/terraform-google-conversion/blob/master/vendor/google.golang.org/api/sqladmin/v1beta4/sqladmin-gen.go#L305-L342) and [TPG](https://github.com/terraform-providers/terraform-provider-google/blob/master/vendor/google.golang.org/api/sqladmin/v1beta4/sqladmin-gen.go#L363-L405)

xingao267:
The go compile error is due to new errors.Unwrap() function introduced in go 1.13. See https://blog.golang.org/go1.13-errors. And the container used in cloud build has go 1.12.8.

xingao267:
Now it's showing the errors I would like to show :)

xingao267:
@danawillow might need your help here. Please see the errors I mentioned above. And also the error [here](https://github.com/GoogleCloudPlatform/terraform-validator/pull/128#issuecomment-591187453). I think they are related or similar as some kind of null or empty list check should be performed in the `terraform-google-conversion` library. My understanding is those files are also auto-generated from magic-module. So I think some kind of fix should go there. Could you take a look at the error and see if that rings a bell?

danawillow:
I don't have access to the cloud build that's currently failing, so I'll do what I can but you may need to share more details with me for some of the failures.

storage_bucket.go is handwritten, so if you have a change you'd like to make you're more than welcome to propose a PR (or file a bug against us to do so) without needing to touch any of the codegen parts. It does seem fishy that `d.GetOk("retention_policy")` returns true but has nothing inside of it, so it's worth figuring out what that means semantically and whether it's a valid case to handle or whether it should be prevented against elsewhere.

sql_database_instance is a handwritten resource, which is why it doesn't necessarily match the api.yaml. Location is defined in the schema at https://github.com/GoogleCloudPlatform/magic-modules/blob/master/third_party/terraform/resources/resource_sql_database_instance.go.erb#L170.

The bigquery one is failing because `d.Get("default_encryption_configuration")` is returning nil and I guess we assume that it would have been an empty list if it's unset. I don't know enough about how/where the TerraformResourceData gets built to know whether the nil value is expected or a bug. It seems like the opposite of the storage bucket one in a way, but I can't imagine it's any different from any of the other nested objects that we generate support for.

xingao267:
@danawillow  thanks for the help.

Created https://github.com/GoogleCloudPlatform/magic-modules/pull/3181 to fix the bucket one. Looking into the sql_database_instance one.

xingao267:
I think I know why Location is not recognized by TV. This repo also depends on `github.com/terraform-providers/terraform-provider-google` which also needs to bring up, which is still at 20190926.

xingao267:
The dependencies are just so old and I'm getting even more errors when updating them.

xingao267:
@morgante `make build` and `make test` are both fine in my local environment. The Cloud Build job seems to fail because the caller (cloud builder service account) doesn't have permission to the testing project. Could you help take a look?

PR #137, opened 2020-04-29
Add support for Spanner Instance to Terraform Validator.

mmontan:
The PR introduces Spanner Instance objects consistent with the Public API representation of Spanner Instances auto-generated by Magic Modules. This is not exactly compatible with the existing CAI export of the same resource.

Validator OPA Policies should be compatible with both representation of the resource, if they need to be shared across monitoring and enforcement.


morgante:
@mmontan FYI this looks fine but I want to figure out why tests are failing before merging.

PR #148, opened 2020-08-01
Add policy change detection capabilities.

cledoux:
A generalized approach to detecting IAMPolicy changes on Assets.

In this approach, we create two sets of Assets: before and after. The after Assets is the currently generated ReadPlannedAssets. In the before, we parse the plan's resource_changes to create tfplan.Assets based on the current rather than planned state.

To detect policy changes, we simply compare the sets of Assets and record when

1. New policies are created.
2. Policies or resources are deleted.
3. Assets with the same name have different policies.

### DIFF
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -1,0 +1,102 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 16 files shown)


policy/policy.go
```diff
@@ -0,0 +1,102 @@
...
+		return nil, errors.Wrap(err, "converting tfplan to CAI assets")
+	}
+
+	return extractOverlay(before, after), nil
+}
+
+// extractOverlay compares the current and planned assets to determine
+// what policies have changed.
+func extractOverlay(before, after []google.Asset) Overlay {
+	emptyPolicy := func() *google.IAMPolicy {return &google.IAMPolicy{Bindings:[]google.IAMBinding{}}}
```
morgante, line 57: Any reason to have this defined inline?

Issue #158, opened 2020-11-23
Validator and TGC are out of sync with the providers


daniel-cit:
Magic-modules is the shared source code base for both Terraform-google-conversion and terraform-provider-google.

Terraform-validator and Terraform-google-conversion are out of sync with current the version of terraform-provider-google. The last sync between then was in PR https://github.com/GoogleCloudPlatform/terraform-validator/pull/129 merged in 2020-03-02.

The provider has been evolved this year while Terraform-google-conversion (tgc) and Terraform-validator have not been updated but for:

- A bump in the version of `forseti-security/config-validator`,
- The addition of policy change detection capability,
- Support for Spanner Instance resource,
- some tests fixes.

This lead to some compatibility issues with code generated by the current version of magic-modules.

We have these basic differences between validator/tgc and the provider as of 2020-11-20:

- [x] The Provider uses `go 1.14` while validator and tgc use `go 1.13` https://github.com/GoogleCloudPlatform/terraform-validator/pull/163 https://github.com/GoogleCloudPlatform/terraform-google-conversion/pull/598
- [x] The Provider code moved to `github.com/hashicorp/terraform-provider-google` and  then changed its go package to the new location but is still referenced as `github.com/terraform-providers/terraform-provider-google` in validator. Validator code needs changes to be able to use the newest version of the provider https://github.com/GoogleCloudPlatform/terraform-validator/pull/165
- [x] Validator has not acknowledge that the provider current version is v3 not v1 since it is referencing it by a pseudo-version not a by a real tagged version. https://github.com/GoogleCloudPlatform/terraform-validator/pull/165
- [x] The Provider uses `terraform-plugin-sdk` v2 and validator still uses `terraform-plugin-sdk` v1 https://github.com/GoogleCloudPlatform/terraform-validator/pull/165
- [x] The Provider uses `google.golang.org/api v0.34.0`, tgc and validator use `google.golang.org/api v0.16.0` https://github.com/GoogleCloudPlatform/terraform-validator/pull/165
- [X] Vendoring has been removed from the provider but is still in used in tgc https://github.com/GoogleCloudPlatform/terraform-google-conversion/pull/596
- [x] melinath: Vendoring has been removed from the provider and tgc but is still used in validator https://github.com/GoogleCloudPlatform/terraform-validator/pull/159
[...]

PR #157, opened 2020-11-17
Added support for project organization policy

rdayama-eng:
Corresponding magic module PR

PR #159, opened 2020-12-08
Removed vendored dependencies

melinath:
They are no longer in use on the gcp provider for terraform, and are harder to deal with than non-vendored dependencies, since they require regular non-reviewable PRs.

Related to #158

---

My understanding is that we want to move away from vendored dependencies, but that's not based on anything super specific, other than that vendoring seems to be an older mechanism that has been phased out in the provider code. I'd love to know what your thoughts are!

Because of the drift introduced in #157, this leaves in place the tgc vendored code via a `replace` statement.


morgante:
I'm not sure if this is the best practice or not? For libraries it makes sense, but for a binary having the guarantee of always being able to build if needed seems important.

In either case we definitely would need to address [this drift](https://github.com/GoogleCloudPlatform/terraform-validator/pull/157#issuecomment-735921150) before removing the vendor/ directory.

melinath:
@morgante sorry for the ping; I had briefly assigned you prior to the tests failing. As you mention, there's drift in the repo that makes removing the vendoring at this time infeasible. Based on discussion with the team, it sounds like we would like to remove the vendoring in the long run; however, I'm not sure if that's our decision to make or not?

melinath:
although actually I guess we could remove vendoring and use a `replace` directive to keep the drifted version of tgc for the moment.

morgante:
> Based on discussion with the team, it sounds like we would like to remove the vendoring in the long run; however, I'm not sure if that's our decision to make or not?

It's ultimately a decision for your team. My main question would be how confident you are in being able to rebuild at will without vendoring?

melinath:
@rileykarson My take on this would be that it sounds like go is now sufficiently mature that there's less worry of things suddenly disappearing? But you'd be more familiar with that than I am.

Removing vendoring would definitely have the benefit of preventing manual drift introduced in vendored dependencies.

rileykarson (approved):
Go Modules has a module cache, which means disappearing modules aren't a big deal imo.

rileykarson (approved):
Go Modules has a module cache, which means disappearing modules aren't a big deal imo.

PR #160, opened 2020-12-09
Removed support for terraform 0.11

melinath:
(Note: Of the 969 lines, ~640 are from go_tf_0_11.sum being deleted.)

In terraform 0.12, there was a major change to the structure of terraform plan files; since terraform-validator reads those files, it needed to simultaneously support both the old and new structures. _(Updated note: Specifically, in 0.11 terraform-validator read the tfplan directly; in 0.12 the [tfjson format](https://www.terraform.io/docs/internals/json-format.html) was introduced and is used instead.)_ However, [Terraform 0.11 support in providers has been deprecated](https://www.hashicorp.com/blog/deprecating-terraform-0-11-support-in-terraform-providers) and [0.11 support conflicts with terraform-plugin-sdk-v2](https://github.com/GoogleCloudPlatform/terraform-validator/issues/158#issue-748914579) which terraform-validator will need to have as a dependency going forward.

I've also chosen to remove the version management framework that was in place. There is the chance that Terraform will need to change its file structure again in the future; however, the docs [currently state](https://www.terraform.io/docs/internals/json-format.html):

> The output includes a `format_version` key, which currently has major version zero to indicate that the format is experimental and subject to change. A future version will assign a non-zero major version and make stronger promises about compatibility. We do not anticipate any significant breaking changes to the format before its first major version, however.

Given that the format now contains canonical `format_version` and `terraform_version` keys, it is not clear that the existing implementation for supporting tf 0.11 is the best path forward if and when the tfplan json format changes, and it's not clear how long in the future that change might be. Clearing out this code should give us more flexibility both in the short term and the long term.


README.md
```diff
@@ -4,7 +4,7 @@ This tool is used to validate terraform plans before they are applied. Validatio
 
 **Note**: Using Terraform Validator does _not_ require an active installation of Forseti. Terraform Validator is a self-contained binary.
 
-Note: this tool supports Terraform v0.12 by default. To switch to use Terraform v0.11, please see the section [Terraform v0.11](#terraform-v011).
+Note: this tool supports Terraform v0.12+.
```
slevenick, line 7: Does it work with 0.13 and 0.14 currently? Or do those require extra work

tfgcv/planned_assets.go
```diff
@@ -44,40 +41,14 @@ func ReadPlannedAssets(ctx context.Context, path, project, ancestry string, offl
...
-		}
-
-		// Attempt to pull the project from the provider.
-		if project == "" {
-			project, err = parseProviderProject(plan)
-			if err != nil {
-				return nil, err
-			}
-		}
+	data, err := readTF12Data(path)
```
slevenick, line 44: What happens if the user is still running 0.11? Will this throw an informative error?

slevenick (approved):

README.md
melinath: TF version support basically has to do with the format of the terraform plan JSON format, which hasn't changed significantly since 0.12, as I understand it.

melinath: I've been testing tf validator locally with tf 13.5 without any issues.

cmd/root.go
```diff
@@ -85,7 +81,7 @@ var rootCmd = &cobra.Command{
 	Long: fmt.Sprintf(`Validate terraform plans by converting terraform resources to their Google CAI 
 (Cloud Asset Inventory) format and passing them through Forseti Config Validator.
 
-Supported Terraform versions = %s`, version.AllSupportedVersions()),
+Supported Terraform versions = 0.12`),
```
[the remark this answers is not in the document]
melinath: Thanks! Changed in f2cf705

README.md
melinath: Confirmed to work with terraform 0.14.2 as well (same JSON `format_version`).

tfgcv/planned_assets.go
melinath: In TF 0.11 validator read the plan file directly; in TF 0.12 it reads a json file. When you try to pass in a non-JSON file you get:

```
Error: converting tfplan to CAI assets: Terraform 0.12 support plans only in JSON format, got: .tfplan
```

PR #163, opened 2020-12-10
Upgraded go version to 1.14

melinath:
Related to #158. In order to upgrade to 1.14 I had to move the `vendor/` folder due to 1.14 interpreting a `vendor` folder to mean things are actually vendored. The remaining item in that folder is not actually vendored; it contains some drift compared to the base repository and can't be removed quite yet. (But it will be removed as part of the work on ticket #158.)


melinath:
@slevenick this was pretty similar to another PR so it may have gotten lost - just wanted to make sure you had seen it :-)

slevenick (approved):
Yup definitely missed it!

melinath:
thanks!

PR #164, opened 2020-12-14
Updated to terraform-plugin-sdk/v2

melinath:


melinath:
Ended up not being easy to tease this out from the rest of the necessary upgrades.

PR #165, opened 2020-12-14
Update terraform-google-conversion, terraform-provider-google, and terraform-plugin-sdk

melinath:
terraform-validator has three primary dependencies that need to be updated at the same time:

- terraform-google-conversion
- terraform-provider-google
- terraform-plugin-sdk

I was not able to make a PR that upgraded just one of these, because of a combination of shifting function signatures and conflicting dependencies. However, most of the 30k lines being removed are from removing the vendoring of terraform-google-conversion, which I hope helps.

I've done my best to note in each commit why I did certain changes, but I'm also happy to answer any specific questions. This is a bit of a mess.

Note: github.com/hashicorp/terraform-provider-google/v3 is pegged at a specific commit to get the `go mod tidy` changes to that repo, which won't be present until the next release (i.e. after the holidays). This was necessary to avoid an unnecessary dependency on terraform 0.13, which was causing a dependency conflict.


melinath:
It looks like the integration tests for this are still failing. I would be happy to fix them on this PR but tbh this is already pretty huge; what would you think about merging with failing integration tests and fixing them in a follow-on PR?

slevenick:
> It looks like the integration tests for this are still failing. I would be happy to fix them on this PR but tbh this is already pretty huge; what would you think about merging with failing integration tests and fixing them in a follow-on PR?

Given that they were already failing I don't have a problem with a PR that gets us part of the way there

slevenick (approved):

PR #168, opened 2020-12-16
Fixed integration tests

melinath:
This PR fixes the integration tests. It is related to #158.

It seems like there are a few systemic changes to the terraform-google-conversion repo and a few that are specific to certain resources. For example, the replicationType field on SQL resources is deprecated.

I tried to update things where it seemed like it might make sense; however, for the most part I assumed that we want tests that validate that the current behavior won't break going forward - i.e. that it is currently working correctly. If that's not the case for any of these, I suggest we fix it in separate PRs.

In addition to the fixes, I also made the integration tests output what versions of go and terraform they use.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,8 @@ run-docker:
 	docker run -it -v `pwd`:/terraform-validator -v ${GOOGLE_APPLICATION_CREDENTIALS}:/terraform-validator/credentials.json --entrypoint=/bin/bash --env TEST_PROJECT=${PROJECT_ID} --env TEST_CREDENTIALS=./credentials.json terraform-validator;
 
 test-integration:
+	go version
+	terraform --version
 	go test -v -run=CLI ./test
 
 build-docker:
--- a/testdata/templates/example_compute_instance.json
+++ b/testdata/templates/example_compute_instance.json
@@ -13,7 +13,7 @@
         "displayDevice": {
           "enableDisplay": false
         },
-        "shieldedVmConfig": {
+        "shieldedInstanceConfig": {
           "enableIntegrityMonitoring": false,
           "enableSecureBoot": false,
           "enableVtpm": false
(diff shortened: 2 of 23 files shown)


testdata/templates/example_compute_instance.json
```diff
@@ -13,7 +13,7 @@
         "displayDevice": {
           "enableDisplay": false
         },
-        "shieldedVmConfig": {
+        "shieldedInstanceConfig": {
```
slevenick, line 16: I'm not super familiar with these tests, but it seems like we are using these files for "golden file" testing? I'm curious how we expect them to be updated as the provider adds support for new fields.

slevenick (approved):
I'd be curious how we want to handle updates to these files in the future. Overall I'm not a huge fan of this type of testing

Makefile
```diff
@@ -13,6 +13,8 @@ run-docker:
 	docker run -it -v `pwd`:/terraform-validator -v ${GOOGLE_APPLICATION_CREDENTIALS}:/terraform-validator/credentials.json --entrypoint=/bin/bash --env TEST_PROJECT=${PROJECT_ID} --env TEST_CREDENTIALS=./credentials.json terraform-validator;
 
 test-integration:
+	go version
```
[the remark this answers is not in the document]
melinath: I wasn't able to figure out what go version and terraform version were being used in the tests otherwise. It ended up not being relevant in this case, but I would rather have those values easily available for the future.

testdata/templates/example_compute_instance.json
melinath: yeah, I hear you on this not being an ideal way of doing tests. I'm not super familiar with them either / not sure how often they'll change - except to say that, in the last 9 or so months, there were about three resource changes that only broke these tests. (There were also resource changes that caused other types of issues...) Definitely into talking about another way to handle this going forward.

c2thorn (approved):
+1 to Sam's sentiments, it'd be nice to do keep these in sync in a more automated way

c2thorn:
Didn't refresh, already merged 🤦

melinath:
Sorry @c2thorn I assigned you but then thought you were probably gone already 😬

PR #172, opened 2021-01-26
Merge project IAM policy bindings

melinath:
This PR sets up merging of project IAM policy bindings from GCP into the local bindings prior to validation. This means that validation and other functionality (like simulation) will be able to reference the full IAM policy, even if the user is using member or binding resources in their terraform.

You can test this manually by making a file locally like this:

```
resource "google_project" "my_project" {
  name       = "Whatever you want here"
  project_id = "<your test project ID>"
  org_id     = "whatever you want here"
}


resource "google_project_iam_binding" "project" {
  project = "<your test project ID>"
  role    = "roles/viewer"

  members = [
    "user:some-user-email",
  ]
}
```

Once you have this file in a directory, run:
```
cd path/to/terraform-validator
make build
cd path/to/test/dir
terraform plan --out=tfplan.tfplan 
terraform show -json tfplan.tfplan > tfplan.json
path/to/terraform-validator/bin/terraform-validator convert tfplan.json
```

Without these changes, the output will only contain the binding defined in terraform; with these changes, it will also contain your project's full iam policy.

I would like to have automated tests for this but am unsure how to go about implementing them. If you have any advice I would love to hear it!

Related to issue https://github.com/hashicorp/terraform-provider-google/issues/7797. Note: This PR only implements project IAM policy related functionality. Follow-up PRs will build on this for other IAM resources.

### DIFF
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -49,6 +49,7 @@ func (nonImplementedResourceData) HasChange(string) bool             { return fa
 func (nonImplementedResourceData) Set(string, interface{}) error     { return nil }
 func (nonImplementedResourceData) SetId(string)                      {}
 func (nonImplementedResourceData) GetProviderMeta(interface{}) error { return nil }
+func (nonImplementedResourceData) Timeout(key string) time.Duration  { return time.Duration(1) }
 
 // Asset contains the resource data and metadata in the same format as
 // Google CAI (Cloud Asset Inventory).
@@ -195,6 +196,31 @@ func (c *Converter) AddResource(r TerraformResource) error {
 		}
 
 		key := converted.Type + converted.Name
+
+		// The existence of a newIamUpdater function signals that this tf
+		// resource needs to be merged with existing remote IAM data to be useful.
+		// This will be run once, for the first IAM policy encountered.
+		if mapper.newIamUpdater != nil {
+			if _, exists := c.assets[key]; !exists {
+				updater, _ := mapper.newIamUpdater(data, c.cfg)
+				iam_policy, _ := updater.GetResourceIamPolicy()
+				var bindings []IAMBinding
+				for _, b := range iam_policy.Bindings {
+					bindings = append(bindings, IAMBinding{
+						Role:    b.Role,
+						Members: b.Members,
+					})
+				}
+				remoteAsset := Asset{
+					Name: converted.Name,
+					Type: converted.Type,
+					IAMPolicy: &IAMPolicy{
+						Bindings: bindings,
+					},
+				}
+				c.assets[key] = remoteAsset
+			}
+		}
 
 		if existing, exists := c.assets[key]; exists {
 			// The existance of a merge function signals that this tf resource maps to a
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -145,6 +145,7 @@ func NewConverter(ctx context.Context, ancestryManager ancestrymanager.AncestryM
 		schema:          provider.Provider(),
 		mapperFuncs:     mappers(),
 		cfg:             cfg,
+		offline:         offline,
 		ancestryManager: ancestryManager,
 		assets:          make(map[string]Asset),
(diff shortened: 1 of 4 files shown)


converters/google/convert.go
```diff
@@ -200,7 +205,34 @@ func (c *Converter) AddResource(r TerraformResource) error {
 			// The existance of a merge function signals that this tf resource maps to a
 			// patching operation on an API resource.
 			if mapper.merge != nil {
-				converted = mapper.merge(existing.converterAsset, converted)
+				converterAsset := existing.converterAsset
+
+				// The existence of a newIamUpdater function signals that this tf
+				// resource needs to be merged with existing remote IAM data to be useful.
+				// We pre-merge that data prior to runner mapper.merge.
+				if mapper.newIamUpdater != nil && !c.offline && existing.IAMPolicy == nil {
```
slevenick, line 213: I'm imagining that eventually we will want similar functionality for other fine-grained resources like [router_nat](https://github.com/hashicorp/terraform-provider-google/blob/master/google/resource_compute_router_nat.go). 
There are several resources that operate this way, where they are essentially sub-resources that are managed individually.

If this functionality for other fine-grained resources is something we want to support in the future it may make sense to move this logic into an IAM specific block or hide it somewhere in the merge or a new fine-grained merge

converters/google/convert.go
```diff
@@ -200,7 +205,34 @@ func (c *Converter) AddResource(r TerraformResource) error {
...
+							Role:    b.Role,
+							Members: b.Members,
+						})
+					}
+					converterAsset.IAMPolicy = &converter.IAMPolicy{
+						Bindings: bindings,
+					}
+				}
+
+				converted = mapper.merge(converterAsset, converted)
```
slevenick, line 235: Does this handle bindings in a TF config that will remove a similar binding on the IAM policy?

converters/google/convert.go
melinath: Interesting - good to know that this might have other use cases.

This solution is definitely a little hacky - I can take a stab at a slightly more factored-out approach. I had considered making it part of the mapper.merge() function - that just means that we need to pass in a converter.Config so that the merge function can make calls to the API. I wasn't sure it was worth it for just IAM but if there will be other resources that need the functionality it's probably a good idea.

converters/google/convert.go
melinath: hm, yeah, there's a good chance the merging logic needs to be refactored to account for deleting assets.

melinath:
Thanks Sam - I'll work on incorporating those points into terraform-google-conversion and then update this PR.

converters/google/convert.go
melinath: Some additional rambling on this point:

Historically, terraform-validator didn't need to care about the diff between the current vs planned assets, because it just needed to make sure the planned state was valid. So if something wasn't present in planned assets, it wouldn't get validated, and that was probably fine, because it would be deleted.

So, the bug we would be introducing by not properly handling deleted iam assets is that validation could get tripped up with something being invalid that has actually already been removed in the plan. For simulator (the root feature request) they would never be able to tell when something had been removed - maybe even worse than the situation they're in now.

melinath: I don't think there's a way to resolve this without making terraform-validator diff-aware - if we only look at the planned resources there's no way to know what Bindings need to be deleted, for example. There is existing work from @cledoux to add a ReadCurrentAssets function in order to be able to diff; however, it functions by reading the current state and planned state separately by diffing them, which wouldn't allow us to access the diff data while adding a single resource (so it doesn't help us handle fine-grained resources.)

I think it might be worthwhile to refactor the system to work with resource changes instead so that each resource being processed has access to both the previous and current state and can understand what changes have occurred.

melinath:
I'm closing this PR as it is not going to be possible to get from here to a solution. I put a design doc laying out the follow-up steps at go/terraform-validator-full-iam

PR #174, opened 2021-02-08
Moved ResourceData faking to converters/google/

melinath:
FakeResourceData provides a central struct that has a simpler structure and more explicitly states its purpose - i.e. faking a ResourceData struct, rather than being a generic field getter or being a Resource (an overloaded term in this context). This is in converters/google instead of tfplan to better separate concerns; tfplan can focus on reading a terraform plan, and converters/google can focus on interfacing between terraform-google-conversion/config-connector.

I was also able to remove Fullpath (provided unused functionality) and newFieldGetter (completely unused.)

This is preparatory work to allow the next step to happen: creating FakeResourceData based on resource change data instead of resource state data. This conversion will need to happen inside the converters/google package because it will happen _after_ determining whether a change is a create, update, or delete, which will be inside the converter as part of the AddResourceChange function.


converters/google/fake_resource_data.go
```diff
@@ -230,3 +217,85 @@ func addrToSchema(addr []string, schemaMap map[string]*schema.Schema) []*schema.
 
 	return result
 }
+
+// attributes function takes json parsed JSON object (value param) and fill map[string]string with it's
```
melinath, line 221: This was moved wholesale from tfplan/json_plan.go.

converters/google/fake_resource_data.go
```diff
@@ -12,85 +12,57 @@
...
 	"strconv"
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
-// newFieldGetter accepts a resource schema map (field name --> Schema),
-// a state instance, and a diff and returns a type with field getter methods.
-func newFieldGetter(sch map[string]*schema.Schema, state *terraform.InstanceState, diff *terraform.InstanceDiff) *fieldGetter {
```
melinath, line 27: This was unused.

converters/google/fake_resource_data.go
```diff
@@ -12,85 +12,52 @@
...
-	schema map[string]*schema.Schema
-	state  *terraform.InstanceState
+// Kind returns the type of resource (i.e. "google_storage_bucket").
+func (d *FakeResourceData) Kind() string {
+	return d.kind
 }
 
 // Id returns the ID of the resource from state.
-func (g *fieldGetter) Id() string {
-	if g.state != nil {
```
melinath, line 72: state was never actually set, so it was always nil. I removed it and this conditional for now; we can always implement this correctly later if needed.

slevenick (approved):

PR #176, opened 2021-02-09
Switched to reading resource changes instead of planned resource values

melinath:
The goal of this PR is to start operating on ResourceChanges, as the next step towards having the converter be able to convert and merge deletions of IAM resources. For this step, the focus is on changing just the data format passed to the converter and leaving the exposed API intact – specifically, maintaining the compatibility described [here](https://docs.google.com/document/d/1_FDdCuqs3hpaceZRXnKfJJ0VgdsJzuKYsmOEA0AZIi8/edit#heading=h.gijrsdjw51q5). ResourceChange reading is simpler because [the data is a simple list instead of a nested structure](https://www.terraform.io/docs/internals/json-format.html#plan-representation).

This is related to https://github.com/hashicorp/terraform-provider-google/issues/7797

### DIFF
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -181,8 +181,9 @@ func (c *Converter) Schemas() map[string]*schema.Resource {
 
 // AddResource converts a terraform resource and stores the converted asset.
 func (c *Converter) AddResource(rc *tfplan.ResourceChange) error {
-	// Compatibility shim: silently do nothing with deleted resources.
-	if len(rc.Change.Actions) == 1 && rc.Change.Actions[0] == "delete" {
+	// Compatibility shim: silently do nothing if this resource isn't
+	// being added or changed.
+	if !(rc.IsCreate() || rc.IsUpdate() || rc.IsDeleteCreate()) {
 		return nil
 	}
 
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -24,6 +24,7 @@ import (
 	provider "github.com/hashicorp/terraform-provider-google/v3/google"
 	"github.com/pkg/errors"
 
+	"github.com/GoogleCloudPlatform/terraform-validator/tfplan"
 	converter "github.com/GoogleCloudPlatform/terraform-google-conversion/google"
 	"github.com/GoogleCloudPlatform/terraform-validator/ancestrymanager"
 )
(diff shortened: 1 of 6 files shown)


converters/google/convert.go
```diff
@@ -179,7 +180,30 @@ func (c *Converter) Schemas() map[string]*schema.Resource {
 }
 
 // AddResource converts a terraform resource and stores the converted asset.
-func (c *Converter) AddResource(r TerraformResource) error {
+func (c *Converter) AddResource(rc *tfplan.ResourceChange) error {
+	// Compatibility shim: silently do nothing with deleted resources.
+	if len(rc.Change.Actions) == 1 && rc.Change.Actions[0] == "delete" {
```
rileykarson, line 185: should we add an `isDelete` helper similar to `isCreate`?

tfplan/json_plan.go
```diff
@@ -59,111 +50,27 @@ type jsonResourceChange struct {
...
-type jsonResource struct {
-	Module       string                 `json:"module"`
-	Name         string                 `json:"name"`
-	Address      string                 `json:"address"`
-	Mode         string                 `json:"mode"`
-	Kind         string                 `json:"type"`
-	ProviderName string                 `json:"provider_name"`
-	Values       map[string]interface{} `json:"values"`
+// ComposeTF12Resources is a thin wrapper around ReadResourceChanges as a compatibility shim.
+func ComposeTF12Resources(data []byte, schemas map[string]*schema.Resource) ([]ResourceChange, error) {
```
rileykarson, line 63: I forget- are we able to change this argument to be named `_`?

rileykarson, line 63: This is something we can eliminate and we just don't want to do that in this PR right?

tfplan/json_plan.go
```diff
@@ -28,19 +25,13 @@ const maxChildModuleLevel = 10000
 // jsonPlan structure used to parse Terraform 12 plan exported in json format by 'terraform show -json ./binary_plan.tfplan' command.
```
rileykarson, line 25: L23 (`maxChildModuleLevel`) is now unused, right?

rileykarson (approved):

tfplan/json_plan.go
melinath: Deleting this would cause an incompatibility as described [here](https://docs.google.com/document/d/1_FDdCuqs3hpaceZRXnKfJJ0VgdsJzuKYsmOEA0AZIi8/edit#heading=h.gijrsdjw51q5). TLDR we would need to update calling code prior to removing this; removing this function would be a breaking change. (But yes, we can eliminate it after the calling code is updated.)

tfplan/json_plan.go
melinath: yep! removed in ec57ba01

tfplan/json_plan.go
melinath: signs point to yes; changed in daf0333d

converters/google/convert.go
melinath: yes, but I think that will fit better in one of the future PRs.

melinath: After thinking about this a bit more I think I need to flip this around, since this gate would let through things that we don't necessarily want to handle (specifically `["no-op"]`, `["read"]`, and `["create", "delete"]`).

melinath: Changed this in b3f75001. I left `IsDelete` for a future PR since we won't be using it yet.

PR #177, opened 2021-02-09
Fully moved responsibility for iterating over and filtering resource changes to converter

melinath:
This change allows the converter to filter out unsupported resource changes (this was partially handled by `tfplan.jsonToResources` prior to #176) and to have a specific internal function that is only responsible for adding a resource change that is known to be a create or update to the converter's internal asset map (which means we will be able to have a separate internal function to handle merging in deletes, and a separate internal function for pre-fetching things like IAM.)

This is slightly different than the design doc, which called for detecting create&update vs delete inside AddResourceChange; however, I think this is a better separation of concerns.

### DIFF
--- a/converters/google/convert_test.go
+++ b/converters/google/convert_test.go
@@ -15,11 +15,33 @@
 package google
 
 import (
+	"context"
 	"sort"
 	"testing"
 
+	"github.com/GoogleCloudPlatform/terraform-validator/ancestrymanager"
+	"github.com/GoogleCloudPlatform/terraform-validator/tfplan"
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 )
+
+const testProject = "test-project"
+
+func newTestConverter() (*Converter, error) {
+	ctx := context.Background()
+	ancestry := ""
+	project := testProject
+	offline := true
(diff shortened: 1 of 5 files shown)


converters/google/convert_test.go
```diff
@@ -47,3 +69,144 @@ func TestSortByName(t *testing.T) {
...
+		Change: tfplan.ChangeRepresentation{
+			Actions: []string{"change"},
+			Before:  nil,
+			After:   nil,
+		},
+	}
+	c, err := newTestConverter()
+	assert.Nil(t, err)
+
+	// fake that this resource is known to the provider; it will never be "supported" by the
```
c2thorn, line 109: just curious, why is that?

c2thorn (approved):

converters/google/convert_test.go
melinath: Since this is a fake terraform resource (`google_unsupported`) we can be confident that it is not going to have support added, so this test shouldn't accidentally break in the future. Does that answer the question?

c2thorn: I see, I thought this was referring to `google_compute_disk`!

PR #180, opened 2021-02-11
Implemented fetch & merge logic on Converter

melinath:
This PR implements fetch & mergeDelete functionality in terraform-validator, with one concrete implementation (Project IAM mergeDelete - see https://github.com/GoogleCloudPlatform/terraform-google-conversion/pull/631).

Local manual testing:

1. Check out this branch of terraform-validator and the https://github.com/GoogleCloudPlatform/terraform-google-conversion/pull/631 branch of terraform-google-conversion
2. Add `replace github.com/GoogleCloudPlatform/terraform-google-conversion v0.0.0-20210105223515-0824b0da180b => /path/to/terraform-google-conversion` to go.mod
3. Run `make build` inside terraform-validator. This will put a binary in `terraform-validator/bin`.
4. Create a terraform file that adds a binding or member to your test project.
5. Apply the file.
6. Remove the binding / member from the file.
7. Run `terraform plan --out tfplan.tfplan`
8. Run `terraform show -json ./tfplan.tfplan > ./tfplan.json`
9. Run `/path/to/terraform-validator/bin/terraform-validator convert ./tfplan.json`

The output of the final command is JSON. You can inspect it to see that your IAM policy (minus the binding / member you removed) is present, despite not being in terraform. You can re-run this process starting at step 3 (but from the terraform-validator master branch) to see the previous behavior (which will give you an empty iam policy.)

~Regarding automated testing: terraform-google-conversion has automated tests of the merge behavior for deletes, but there are no integration tests in this PR for the fetch behavior. I'm not sure how to implement them in the context of terraform-validator, since it isn't set up for VCR tests as far as I know. Please let me know if you have suggestions on how I could implement something!~ See comment below re: automated tests.

Related to https://github.com/hashicorp/terraform-provider-google/issues/7797

This is a breaking change; it looks like the last release was 2021-01-21, right before the changes for supporting IAM better started coming through, so I think we don't need to do a release prior to merging this PR.


converters/google/fake_resource_data.go
```diff
@@ -77,6 +78,7 @@ func (d *FakeResourceData) HasChange(string) bool             { return false }
 func (d *FakeResourceData) Set(string, interface{}) error     { return nil }
 func (d *FakeResourceData) SetId(string)                      {}
 func (d *FakeResourceData) GetProviderMeta(interface{}) error { return nil }
+func (d *FakeResourceData) Timeout(key string) time.Duration  { return time.Duration(1) }
```
melinath, line 81: required for compatibility with the latest versions of terraform-google-conversion

melinath:
Note: the build is failing because it's not using an up-to-date version of terraform-google-conversion. Once the related TGC changes are merged, I'll update the version in this PR.

melinath:
Integration tests are failing due to the changes in this PR - basically, because IAM resources now include the remote data when converted, so the output is no longer as expected. I think the right thing to do would be to update the tests to match - but any alterations to the remote policies would then need to be reflected in changes to the tests. Still, that might be the best option in the short term; in the longer term we can set up VCR for the integration tests to decouple them better.

Does that sound reasonable?

converters/google/convert.go
```diff
@@ -180,19 +187,78 @@ func (c *Converter) AddResourceChanges(changes []tfplan.ResourceChange) error {
...
-			if err := c.addCreateOrUpdate(rc); err != nil {
-				if errors.Cause(err) == ErrDuplicateAsset {
-					glog.Warningf("adding resource change: %v", err)
-				} else {
-					return errors.Wrapf(err, "adding resource change")
-				}
+			createOrUpdates = append(createOrUpdates, rc)
+		} else if rc.IsDelete() {
+			if err := c.addDelete(rc); err != nil {
+				return errors.Wrapf(err, "adding resource deletion")
```
rileykarson, line 193:
```suggestion
				return fmt.Errorf("adding resource deletion %w", err)
```

nit: `errors.Wrapf` is largely obviated by a core language feature as of `1.13`

melinath:
This is a breaking change; it looks like the last release was 2021-01-21, right before the changes for supporting IAM better started coming through, so I think we don't need to do a release prior to merging this PR.

converters/google/convert.go
```diff
@@ -212,17 +278,26 @@ func (c *Converter) addCreateOrUpdate(rc tfplan.ResourceChange) error {
...
+		var existingConverterAsset *converter.Asset
+		if !exists && mapper.fetch != nil{
+			asset, err := mapper.fetch(&rd, c.cfg)
+			existingConverterAsset = &asset
+			if err != nil {
+				return errors.Wrap(err, "fetching asset")
+			}
+		} else if exists {
+			existingConverterAsset = &existing.converterAsset
+		}
```
rileykarson, line 291: Is there any reason we can't structure this like the following? Same logic (I think) but doesn't need to extra `!exists` check and the variables are in scope less long.

```suggestion
		var existingConverterAsset *converter.Asset
		if existing, exists := c.assets[key]; exists {
			existingConverterAsset = &existing.converterAsset
		} else if mapper.fetch != nil{
			asset, err := mapper.fetch(&rd, c.cfg)
			if err != nil {
				return errors.Wrap(err, "fetching asset")
			}

			existingConverterAsset = &asset
		}
```

rileykarson (commented):
Would it make sense to switch the IAM tests to match part of the output instead of the whole output maybe? It'd be nicer to make them stable within this PR.

converters/google/convert.go
melinath: ah, great!

melinath: Replaced this and a few other instances in this file in 41dd2648

melinath: The one remaining usage is making the error a specific type, which seems like a different thing than what fmt.Errorf does?

converters/google/convert.go
melinath: I don't remember how I ended up with that particular flow but I think your version is cleaner & does the same thing (plus moving setting of `existingConverterAsset = &asset` until after the error check is better.)

melinath:
@rileykarson re: the IAM tests - I could certainly limit them to just checking that the things that were expected to be added were actually added. However, that would then mean that there's nothing checking whether the functionality introduced by this PR is working as intended. If you would prefer to put that off until we can get VCR tests set up, that's okay with me.

converters/google/convert.go
```diff
@@ -212,17 +278,26 @@ func (c *Converter) addCreateOrUpdate(rc tfplan.ResourceChange) error {
...
+
+			existingConverterAsset = &asset
+		}
+
+		if existingConverterAsset != nil {
+			if mapper.mergeCreateUpdate == nil {
 				// If a merge function does not exist ignore the asset and return
 				// a checkable error.
 				return errors.Wrapf(ErrDuplicateAsset, "asset type %s: asset name %s",
 					converted.Type, converted.Name)
```
rileykarson, line 298:
```suggestion
				return fmt.Errorf("asset type %s: asset name %s %w", converted.Type, converted.Name, ErrDuplicateAsset)
```

converters/google/convert.go
rileykarson: https://github.com/GoogleCloudPlatform/terraform-validator/pull/180/files#r577225621 should do the same thing using the core interface

rileykarson (approved):
They'd verify the general working-ness of the IAM resources but not the merging behaviour specifically, right? Up to you!

converters/google/convert.go
melinath: ah, cool, wasn't sure if that would work 👍

melinath:
@rileykarson I ended up making the project iam tests only check that the expected values are present, since the integration tests can be run against arbitrary GCP projects. I'm not sure if there's an elegant / repeatable way to test merge behavior specifically in the current framework, but I don't want to add that as part of this PR since it's already ballooned a bit.

The changes to get tests running actually ended up being a bit more extensive than I expected; if you have time to have a look and make sure they seem reasonable / in line with best Go practices that would be great.

rileykarson (approved):
LGTM

We could verify using a less busy IAM resource than a project. https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/pubsub_topic_iam is pretty cheap to provision?

melinath:
okay - I'll keep that in mind when I add pubsub_topic support :-) Thanks!

PR #191, opened 2021-03-02
Added note about how to check if a resource is already supported by tgc

melinath:
This is a thing that is not obvious to people not familiar with the system. There are a lot of resources supported by TGC but not validator, so we should point people to them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ Some Terraform resources are fully generated, whereas some resources are hand wr
 
 Adding support for a new resource follows these steps:
 
-0. If the resource is already [supported by terraform-google-conversion](https://github.com/GoogleCloudPlatform/terraform-google-conversion/tree/master/google) you can skip to step 6. You may be need to upgrade the terraform-google-conversion dependency in terraform-validator's go.mod file.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -147,6 +147,7 @@ Some Terraform resources are fully generated, whereas some resources are hand wr
 
 Adding support for a new resource follows these steps:
 
+0. If the resource is already [supported by terraform-google-conversion](https://github.com/GoogleCloudPlatform/terraform-google-conversion/tree/master/google) you can skip to step 6. You may be need to upgrade the terraform-google-conversion dependency in terraform-validator's go.mod file.
```
c2thorn, line 150: Is this helpful? If not, cool!
```suggestion
0. If the resource is already [supported by terraform-google-conversion](https://github.com/GoogleCloudPlatform/terraform-google-conversion/tree/master/google) you can skip to step 6. You may be need to upgrade the terraform-google-conversion dependency in terraform-validator's [go.mod file](https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/go.mod).
```

c2thorn (approved):

README.md
melinath: I was going back and forth on it :-) let's do it!

PR #195, opened 2021-03-11
support multiple assets instances

ericyz:


ericyz:
What is the failure message says?

ericyz:
Okay, I can see it from Cloud Build. 

github.com/GoogleCloudPlatform/terraform-google-conversion/google.expandNetworkInterfaces(0x41bc0a0, 0xc000f40ff0, 0xc00023a000, 0x0, 0x0, 0x1, 0x1, 0x1)
	/builder/home/go/pkg/mod/github.com/!google!cloud!platform/terraform-google-conversion@v0.0.0-20210311162721-d97a3783d011/google/compute_instance_helpers.go:248 +0xcaa
github.com/GoogleCloudPlatform/terraform-google-conversion/google.expandComputeInstance(0xc000c35558, 0x6, 0x41bc0a0, 0xc000f40ff0, 0xc00023a000, 0xc000c35558, 0x6, 0x0)
	/builder/home/go/pkg/mod/github.com/!google!cloud!platform/terraform-google-conversion@v0.0.0-20210311162721-d97a3783d011/google/compute_instance.go:122 +0x731
github.com/GoogleCloudPlatform/terraform-google-conversion/google.GetComputeInstanceApiObject(0x41bc0a0, 0xc000f40ff0, 0xc00023a000, 0x3ae6d91, 0x4f, 0xc000058ea0)
	/builder/home/go/pkg/mod/github.com/!google!cloud!platform/terraform-google-conversion@v0.0.0-20210311162721-d97a3783d011/google/compute_instance.go:47 +0xa1
github.com/GoogleCloudPlatform/terraform-google-conversion/google.GetComputeInstanceCaiObject(0x41bc0a0, 0xc000f40ff0, 0xc00023a000, 0x17, 0xc0005816f0, 0xc000f550a0, 0xc000052b40, 0x17)
	/builder/home/go/pkg/mod/github.com/!google!cloud!platform/terraform-google-conversion@v0.0.0-20210311162721-d97a3783d011/google/compute_instance.go:25 +0xad

It seems the error message comes from the type conversion. I guess it is caused by a recent change
https://github.com/GoogleCloudPlatform/magic-modules/blob/06b0fd539a48b8fd4040321e7b0db2ecfb5cf970/mmv1/third_party/terraform/utils/compute_instance_helpers.go.erb#L258

melinath:
Yeah that does seem to be the case. I can reproduce this failure locally as well. I think you could resolve this by updating the `testdata/templates/full_compute_instance.*` files to include the new fields.

ericyz:
Hi @melinath , it seems the data model (computeBeta.NetworkInterface) reference here is outdated. The dictionary doesn't include "nic_type". What would a good way you suggest to update the reference?

melinath:
Hi, Eric - I looked into this more closely and I was wrong; we can't fix this by updating the testdata files (or at least not just that way.) I think you're right that it may need to be fixed in magic modules. I'll see if I can get something together today.

melinath:
Okay, I have a PR to fix magic modules. I've tested the change locally and it seems like we shouldn't need to alter the test files at all - having `null` values in the json is expected; they just need to be treated as strings instead of interfaces during expansion.

melinath (approved):
TLDR: in order to get the tests to pass, we need to have an updated version of terraform-google-conversion _and_ an updated version of terraform-google-provider - but we can't get an updated version of terraform-google-provider easily because we removed `/v3` markers (see https://github.com/hashicorp/terraform-provider-google/pull/8538). I'll look into fixing that separately, since it's a bit of a mess. In the meantime: this does seem to be working as intended, so I'm approving and merging :-) Thank you!

melinath:
upstream change: https://github.com/GoogleCloudPlatform/magic-modules/pull/4567

PR #198, opened 2021-03-16
to support filestore

ericyz:


melinath (approved):
The issue with nic_type is fixed on master; I've verified locally that these tests pass with master merged in.

PR #199, opened 2021-03-17
Updated tpg to a commit patched to provide 3.60 with /v3

melinath:
We need to upgrade terraform-provider-google because we need the schema changes introduced in https://github.com/GoogleCloudPlatform/magic-modules/pull/4443.

Changing to a specific commit is necessary because terraform-provider-google releases are no longer importable as go modules by default. See https://github.com/hashicorp/terraform-provider-google/pull/8538 for the reasoning.

This is intended as a quick fix until we can get a more permanent solution in place for giving validator access to the terraform schema.


c2thorn (approved):

melinath:
See related discussion here: https://github.com/GoogleCloudPlatform/terraform-validator/pull/195

Issue #196, opened 2021-03-11
Validator fails with two project with billing budget and random project-id


daniel-cit:
Terraform-validator fails with two project with billing budget and random project-id in initial plan.  

Affected version:  **`2021-03-05`**
Terraform version used to generate the plan: **`v0.13.6`**

Version **`2021-03-05`** of terraform-validator fails on a tfplan with the creation of two project with billing budget when the project id uses a random_id resource.

Sample terraform configuration to reproduce the issue:

```terraform
variable "org_id" { type = string }

variable "billing_account" { type = string }

resource "random_id" "random_project_id_suffix" { byte_length = 2 }

resource "google_project" "my_project1" {
  name            = "My Project1"
  project_id      = "your-project-id1-${random_id.random_project_id_suffix.hex}"
  org_id          = var.org_id
  billing_account = var.billing_account
}

resource "google_project" "my_project2" {
  name            = "My Project2"
  project_id      = "your-project-id2-${random_id.random_project_id_suffix.hex}"
  org_id          = var.org_id
  billing_account = var.billing_account
}

resource "google_billing_budget" "budget1" {
  billing_account = var.billing_account
  display_name    = "Example Billing Budget 1"
  amount {
    specified_amount {
      currency_code = "USD"
      units         = "100"
    }
  }
  budget_filter {
    projects = ["projects/${google_project.my_project1.number}"]
  }
  threshold_rules { threshold_percent = 0.5 }
}

resource "google_billing_budget" "budget2" {
  billing_account = var.billing_account
  display_name    = "Example Billing Budget 2"
  amount {
    specified_amount {
      currency_code = "USD"
      units         = "100"
    }
  }
  budget_filter {
    projects = ["projects/${google_project.my_project2.number}"]
  }
  threshold_rules { threshold_percent = 0.5 }
}
```

Running terraform-validator on generated tfplan:

**Build version: `2021-03-05`**

```console
me@example.com:~$ terraform-validator validate two_projects_with_budget.json --policy-path=./policy-library/  --project=prj-c-logging-f077 --verbose
```

```console
2021/03/10 22:26:57 [INFO] Authenticating using DefaultClient...
[...]

PR #200, opened 2021-03-18
returns ErrDuplicateAsset wrapped by a new error so that errors.Cause works

daniel-cit:
This PR Closes https://github.com/GoogleCloudPlatform/terraform-validator/issues/196

TODO:
- [x] Add unit test for the issue

The converter does a check by the cause of an error and just logs a warning if the error cause is `ErrDuplicateAsset `

https://github.com/GoogleCloudPlatform/terraform-validator/blob/6a9552f3f4c079b6c6abd1a53fc4f51f477beb8e/converters/google/convert.go#L203

```go
if errors.Cause(err) == ErrDuplicateAsset {
	glog.Warningf("adding resource change: %v", err)
} else {
``` 

in previous versions, this error was create as:

```go
return errors.Wrapf(ErrDuplicateAsset, "asset type %s: asset name %s", converted.Type, converted.Name)
```

but now it is using:

```go
return fmt.Errorf("asset type %s: asset name %s %w", converted.Type, converted.Name, ErrDuplicateAsset)
```

which does not wrap the error but creates a new error using only the "error message" from the original error.

this PR reverts the code to its previous version

Result with the new version in the test file from the issue(reformatted for clarity):

` ./bin/terraform-validator convert ../two_project_with_budget/tst.json --project=<REDACTED>`

```console
ERROR: logging before flag.Parse: I0318 16:53:12.402048    5683 convert.go:188] unsupported resource: google_billing_budget
ERROR: logging before flag.Parse: I0318 16:53:12.402104    5683 convert.go:188] unsupported resource: google_billing_budget
ERROR: logging before flag.Parse: I0318 16:53:12.402114    5683 convert.go:182] unknown resource: random_id
ERROR: logging before flag.Parse: W0318 16:53:13.708944    5683 convert.go:204] adding resource change: asset type cloudbilling.googleapis.com/ProjectBillingInfo: asset name //cloudbilling.googleapis.com/projects/<REDACTED>/billingInfo: duplicate asset
[
    {
        "name": "//cloudbilling.googleapis.com/projects/<REDACTED>/billingInfo",
        "asset_type": "cloudbilling.googleapis.com/ProjectBillingInfo",
        "ancestry_path": "organization/<REDACTED>/folder/146035178388/folder/220237755881/project/<REDACTED>",
        "resource": {
            "version": "v1",
            "discovery_document_uri": "https://www.googleapis.com/discovery/v1/apis/cloudbilling/v1/rest",
            "discovery_name": "ProjectBillingInfo",
            "parent": "//cloudresourcemanager.googleapis.com/projects/<REDACTED>",
            "data": {
                "billingAccountName": "billingAccounts/<REDACTED>",
                "name": "projects//billingInfo"
            }
        }
    },
    {
        "name": "//cloudresourcemanager.googleapis.com/projects/<REDACTED>",
        "asset_type": "cloudresourcemanager.googleapis.com/Project",
        "ancestry_path": "organization/<REDACTED>/folder/146035178388/folder/220237755881/project/<REDACTED>",
        "resource": {
            "version": "v1",
            "discovery_document_uri": "https://www.googleapis.com/discovery/v1/apis/compute/v1/rest",
            "discovery_name": "Project",
            "parent": "//cloudresourcemanager.googleapis.com/projects/<REDACTED>",
            "data": {
                "name": "My Project2",
                "parent": {
                    "id": "<REDACTED>",
                    "type": "organization"
                }
            }
        }
    }
]
```

### DIFF
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -16,19 +16,20 @@ package google
 
 import (
 	"context"
+	errorssyslib "errors"
 	"fmt"
 	"sort"
 	"time"
 
-	"github.com/hashicorp/terraform-json"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/golang/glog"
+	tfjson "github.com/hashicorp/terraform-json"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	provider "github.com/hashicorp/terraform-provider-google/v3/google"
 	"github.com/pkg/errors"
 
-	"github.com/GoogleCloudPlatform/terraform-validator/tfplan"
 	converter "github.com/GoogleCloudPlatform/terraform-google-conversion/google"
 	"github.com/GoogleCloudPlatform/terraform-validator/ancestrymanager"
+	"github.com/GoogleCloudPlatform/terraform-validator/tfplan"
 )
 
 var ErrDuplicateAsset = errors.New("duplicate asset")
@@ -143,7 +144,7 @@ type Converter struct {
 	mapperFuncs map[string][]mapper
 
 	offline bool
-	cfg *converter.Config
+	cfg     *converter.Config
 
 	// ancestryManager provides a manager to find the ancestry information for a project.
 	ancestryManager ancestrymanager.AncestryManager
@@ -200,7 +201,7 @@ func (c *Converter) AddResourceChanges(changes []*tfjson.ResourceChange) error {
 
 	for _, rc := range createOrUpdates {
 		if err := c.addCreateOrUpdate(rc); err != nil {
-			if errors.Cause(err) == ErrDuplicateAsset {
+			if errorssyslib.Is(err, ErrDuplicateAsset) {
 				glog.Warningf("adding resource change: %v", err)
 			} else {
 				return fmt.Errorf("adding resource create or update %w", err)
@@ -210,7 +211,6 @@ func (c *Converter) AddResourceChanges(changes []*tfjson.ResourceChange) error {
 
 	return nil
 }
-
 
 // For deletions, we only need to handle mappers that support
 // both fetch and mergeDelete. Supporting just one doesn't
@@ -302,7 +302,7 @@ func (c *Converter) addCreateOrUpdate(rc *tfjson.ResourceChange) error {
 				if mapper.mergeCreateUpdate == nil {
 					// If a merge function does not exist ignore the asset and return
 					// a checkable error.
-					return errors.Wrapf(ErrDuplicateAsset, "asset type %s: asset name %s", converted.Type, converted.Name)
+					return fmt.Errorf("asset type %s: asset name %s %w", converted.Type, converted.Name, ErrDuplicateAsset)
 				}
 				converted = mapper.mergeCreateUpdate(*existingConverterAsset, converted)
 			}
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -302,7 +302,7 @@ func (c *Converter) addCreateOrUpdate(rc *tfjson.ResourceChange) error {
 				if mapper.mergeCreateUpdate == nil {
 					// If a merge function does not exist ignore the asset and return
 					// a checkable error.
-					return fmt.Errorf("asset type %s: asset name %s %w", converted.Type, converted.Name, ErrDuplicateAsset)
+					return errors.Wrapf(ErrDuplicateAsset, "asset type %s: asset name %s", converted.Type, converted.Name)
 				}
 				converted = mapper.mergeCreateUpdate(*existingConverterAsset, converted)
 			}
--- a/converters/google/convert_test.go
+++ b/converters/google/convert_test.go
@@ -19,8 +19,8 @@ import (
 	"sort"
 	"testing"
 
-	"github.com/hashicorp/terraform-json"
 	"github.com/GoogleCloudPlatform/terraform-validator/ancestrymanager"
+	tfjson "github.com/hashicorp/terraform-json"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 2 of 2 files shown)


melinath (changes requested):
This LGTM but it would be great to have a test (ideally a unit test if possible) to verify that duplicates don't halt processing.

converters/google/convert.go
```diff
@@ -302,7 +302,7 @@ func (c *Converter) addCreateOrUpdate(rc *tfjson.ResourceChange) error {
 				if mapper.mergeCreateUpdate == nil {
 					// If a merge function does not exist ignore the asset and return
 					// a checkable error.
-					return fmt.Errorf("asset type %s: asset name %s %w", converted.Type, converted.Name, ErrDuplicateAsset)
```
rileykarson, line 305: Shouldn't the `%w` directive have wrapped this? As described at https://go.googlesource.com/proposal/+/master/design/29934-error-values.md#changes-to.

Instead of `errors.Cause` from `github.com/pkg/errors` can we use [`errors.Is` from the system library](https://golang.org/pkg/errors/#example_Is) or is it significantly semantically different?

rileykarson (commented):
Just looking as an FYI, not gonna click any review bits. That's all @melinath.

converters/google/convert.go
daniel-cit: @rileykarson very good question.

After some testing and GitHub exploration we got:

| _Creation_  \\ _Checking_ |  **errors.Is**(github.com/golang/go) | **errors.Cause**(github.com/pkg) |
|---|:-:|:-:|
| **fmt.Errorf**(github.com/golang/go) | OK | **FAIL** |
| **errors.Wrapf**(github.com/pkg) | OK \* | OK |


The  `%w ` directive does wrap the error as indicted in the link, the issue is that:


#### The `fmt.Errorf` + `errors.Is` pair uses a chain of `Unwrap()` calls to navigate the error chain

https://github.com/golang/go/blob/go1.14.15/src/fmt/errors.go#L17

```go
func Errorf(format string, a ...interface{}) error {
	p := newPrinter()
	p.wrapErrs = true
	p.doPrintf(format, a)
	s := string(p.buf)
	var err error
	if p.wrappedErr == nil {
		err = errors.New(s)
	} else {
		err = &wrapError{s, p.wrappedErr}
	}
	p.free()
	return err
}
```


https://github.com/golang/go/blob/go1.14.15/src/errors/wrap.go#L14

```go
func Unwrap(err error) error {
	u, ok := err.(interface {
		Unwrap() error
	})
	if !ok {
		return nil
	}
	return u.Unwrap()
}
```


#### The `errors.Wrapf` + `errors.Cause` pair uses a chain of `Cause()` calls to navigate the error chain

https://github.com/pkg/errors/blob/v0.9.1/errors.go#L201

```go
func Wrapf(err error, format string, args ...interface{}) error {
	if err == nil {
		return nil
	}
	err = &withMessage{
		cause: err,
		msg:   fmt.Sprintf(format, args...),
	}
	return &withStack{
		err,
		callers(),
	}
}
```


https://github.com/pkg/errors/blob/v0.9.1/errors.go#L275
```go
func Cause(err error) error {
	type causer interface {
		Cause() error
	}

	for err != nil {
		cause, ok := err.(causer)
		if !ok {
			break
		}
		err = cause.Cause()
	}
	return err
}
```

#### **\*** The `errors.Wrapf`  + `errors.Is` pair  works because struct `withMessage` implements `Unwrap()` 

https://github.com/pkg/errors/blob/v0.9.1/errors.go#L248

```go
func (w *withMessage) Unwrap() error { return w.cause }
```

Considering this context and that `github.com/golang/go/errors` is not used anywhere in terraform-validator, but we still have three cases of usage of `fmt.Errorf` in the code:

https://github.com/GoogleCloudPlatform/terraform-validator/blob/6cc8af9992499c25aeea698d71d944140712a82c/converters/google/convert.go#L196

https://github.com/GoogleCloudPlatform/terraform-validator/blob/6cc8af9992499c25aeea698d71d944140712a82c/converters/google/convert.go#L206

https://github.com/GoogleCloudPlatform/terraform-validator/blob/6cc8af9992499c25aeea698d71d944140712a82c/converters/google/convert.go#L345

which were introduced as  part of PR https://github.com/GoogleCloudPlatform/terraform-validator/pull/180 (which cuts and restarts the error chain at this point)

I think that:

1) This commit https://github.com/GoogleCloudPlatform/terraform-validator/commit/41dd2648e3c544793bd3501108aa564f83fcb65a cloud be reverted in this Pull Request
2) A new issue could be open to plan how to migrate:

- `errors.Cause()` => `errors.Is()` 
- `errors.Wrapf()` => `fmt.Errorf()`

In the whole terraform-validator at the same time.



@rileykarson @melinath what do you think?

daniel-cit:
Unit test result without fix:
 
```
--- FAIL: TestAddDuplicatedResources (0.00s)
    convert_test.go:347:
                Error Trace:    convert_test.go:347
                Error:          Expected nil, but got: &fmt.wrapError{msg:"adding resource create or update asset type cloudbilling.googleapis.com/ProjectBillingInfo: asset name //cloudbilling.googleapis.com/projects/test-project/billingInfo duplicate asset", err:(*fmt.wrapError)(0xc0001782c0)}
                Test:           TestAddDuplicatedResources
FAIL
FAIL    github.com/GoogleCloudPlatform/terraform-validator/converters/google    0.079s
FAIL
```    

Unit test result with fix:

```
ok      github.com/GoogleCloudPlatform/terraform-validator/converters/google    0.099s
```

converters/google/convert.go
melinath: Thanks for the thorough investigation! It looks like ErrDuplicateAsset is the only case where one of the changes from https://github.com/GoogleCloudPlatform/terraform-validator/commit/41dd2648e3c544793bd3501108aa564f83fcb65a could result in this kind of bug; the other cases are not directly being checked anywhere. Based on the discussion, my preference would be to keep `fmt.Errorf()`and switch the check for ErrNoConversion to use `errors.Is()`. We can then have the conversation about the remaining usages of `errors.Wrapf()` / `errors.Cause()` separately.

daniel-cit: PR updated to use `errors.Is` + `fmt.Errorf`

melinath (approved):
lgtm

Issue #206, opened 2021-03-24
Support validation of organization and folder level resources


melinath:
Currently, asset conversion requires that the asset is a direct descendant of a project. However, not all resources that a person could want to validate are direct descendants of a project. For example, storage bucket iam (which is supported) doesn't have a "project" property, so we have to work around it by relying on the "project" property of the related storage bucket (which works because they're merged together.)

However, for other objects - specifically for this ticket, folder and organization - the requirement of having a related project prevents us from being able to validate them.

We should find a way to support folders and organizations.

melinath:
This will likely be a significant lift, but it would be great to do and potentially unlock other improvements in the validator.

melinath:
This is happening partially because augmentAsset is run when the resource change is added. It should probably be possible to merge all the resource changes first, and augment the assets afterwards. This would also solve the problem where deleting a storage bucket iam policy will fail unless you specify `--project` because it can only find its related project if it's been merged with its related storage bucket already.

melinath:
As part of this refactor, we could also consider grouping changes by underlying CAI asset prior to merging - see https://github.com/GoogleCloudPlatform/terraform-validator/pull/207#discussion_r605841611

melinath:
If we split augmentAsset out from conversion, it would also allow us to implement the augmentAsset step as taking an ancestryManager (instead of having that be a property of the converter struct). Potentially we could even consider moving asset augmentation off the converter altogether.

melinath:
b/169953815

PR #207, opened 2021-03-25
Made storage bucket IAM support merge delete

melinath:
I also added additional / more complete tests for storage bucket IAM. I ran into a bit of confusion due to the issue described in https://github.com/GoogleCloudPlatform/terraform-validator/issues/206; I was able to find a fairly simple workaround for storage bucket IAM for now.

Ultimately, the new tests don't actually verify that merge delete works properly, because that would require being able to fake the results of a mapper.fetch() call, which is outside the scope of this ticket. However, we do have tests in terraform-google-conversion for merge deletion behavior. The code can be manually tested using the following process:

1. Create a terraform file that adds a binding or member to your test project.
2. Apply the file.
3. Remove the binding / member from the file.
4. Run `terraform plan --out tfplan.tfplan`
5. Run `terraform show -json ./tfplan.tfplan > ./tfplan.json`
6. Run `/path/to/terraform-validator/bin/terraform-validator convert ./tfplan.json --project=project_id`

Specifying `--project=project_id` is unfortunately necessary because at the time that deletions are merged, we don't have access to the storage bucket's project. See https://github.com/GoogleCloudPlatform/terraform-validator/issues/206 for more details.

Related to https://github.com/hashicorp/terraform-provider-google/issues/7797


converters/google/mappers.go
```diff
@@ -152,12 +152,16 @@ func mappers() map[string][]mapper {
...
 				mergeCreateUpdate: converter.MergeBucketIamBinding,
+				mergeDelete:       converter.MergeBucketIamBindingDelete,
+				fetch:             converter.FetchBucketIamPolicy,
 			},
 		},
 		"google_storage_bucket_iam_member": {
 			{
 				convert:           converter.GetBucketIamMemberCaiObject,
 				mergeCreateUpdate: converter.MergeBucketIamMember,
+				mergeDelete:       converter.MergeBucketIamMemberDelete,
```
slevenick, line 163: Is there a reason to keep mergeCreateUpdate and mergeDelete separate? I think of the "merge" action as necessarily including both

converters/google/vendor_utils.go
```diff
@@ -41,6 +41,14 @@ func getProject(d converter.TerraformResourceData, config *converter.Config, cai
 		} else {
 			log.Printf("[WARN] Failed to retrieve project_id for %s from resource", cai.Name)
 		}
+	case "storage.googleapis.com/Bucket":
```
slevenick, line 44: Does a bucket's IAM policy have a project ID, and is that project ID actually important?

Normally we need the project for an IAM policy because it is used to construct the URL/self link for the IAM policy, but in the case of storage I don't think that's true

slevenick (approved):
Some questions but nothing blocking

converters/google/mappers.go
melinath: Yes - we do the merge process in three phases.

1. Fetch full IAM policy from GCP.
2. Process all deleted terraform resources for that policy.
3. Process all added / updated terraform resources for that policy.

We could instead have a single merge function that took a list of changes related to a particular iam policy and split the logic internally so developers didn't have to worry about it, but that would require a significant refactor of terraform validator internals. I'm not sure whether it would be an improvement over the current code or not; there are some things I like about the idea. But either way, it's outside of the scope of this ticket.

converters/google/vendor_utils.go
melinath: Storage buckets have a project ID. Storage bucket IAM doesn't, but we can get it from the storage bucket if they've been merged.

The only reason this needs to be here is that terraform-validator currently requires it to exist & errors if it's not present. See #206 for more details.

Issue #216, opened 2021-04-27
terraform-validator crashes when validating google_storage_bucket with google_storage_bucket_iam_member


daniel-cit:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave _+1_ or _me too_ comments; they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If the issue is assigned to a user, that user is claiming responsibility for the issue.

### Terraform Validator version

terraform-validator: v2021-04-07

### Affected Resource(s)

* google_storage_bucket_iam_member
* google_storage_bucket

### Terraform Plan JSON

Gist for the json https://gist.github.com/daniel-cit/607b1251640b6f5f5b11a8b7d8657f9b

Created from this terraform file, which is the Google  Log Export Module example for [Storage+Folder](https://github.com/terraform-google-modules/terraform-google-log-export/tree/master/examples/storage/folder)

```terraform
variable "project_id" {
  description = "The ID of the project in which the log export will be created."
  type        = string
}

variable "parent_resource_id" {
  description = "The ID of the project in which storage bucket destination will be created."
  type        = string
}

resource "random_string" "suffix" {
  length  = 4
  upper   = false
  special = false
}

module "log_export" {
  source                 = "terraform-google-modules/log-export/google"
  destination_uri        = module.destination.destination_uri
  filter                 = "resource.type = gce_instance"
  log_sink_name          = "storage_folder_${random_string.suffix.result}"
  parent_resource_id     = var.parent_resource_id
  parent_resource_type   = "folder"
  unique_writer_identity = true
}

module "destination" {
  source                   = "terraform-google-modules/log-export/google//modules/storage"
  project_id               = var.project_id
  storage_bucket_name      = "storage_folder_${random_string.suffix.result}"
  log_sink_writer_identity = module.log_export.writer_identity
}
```

### Debug Output

https://gist.github.com/daniel-cit/e0f2e3642ac5d83a62d7aea2b2fd74d3

### Expected Behavior

terraform-validator should have validated the plan.

### Actual Behavior

terraform-validator crashes with error:

```
[...]

PR #218, opened 2021-05-17
Log warning for assets that did not return a value for ID field

c2thorn:
Closes https://github.com/GoogleCloudPlatform/terraform-validator/issues/216

Tested with a tfjson plan identical to the one in the issue.

### DIFF
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -259,10 +259,14 @@ func (c *Converter) addDelete(rc *tfjson.ResourceChange) error {
 				existingConverterAsset = &existing.converterAsset
 			} else if !c.offline {
 				asset, err := mapper.Fetch(&rd, c.cfg)
-				if err != nil {
+				if errors.Cause(err) == converter.ErrEmptyIdentityField {
+					glog.Warningf("%s did not return a value for ID field. Skipping asset fetch.", key)
+					existingConverterAsset = nil
+				} else if err != nil {
 					return errors.Wrap(err, "fetching asset")
+				} else {
+					existingConverterAsset = &asset
 				}
-				existingConverterAsset = &asset
 				if existingConverterAsset != nil {
 					converted = mapper.MergeDelete(*existingConverterAsset, converted)
 					augmented, err := c.augmentAsset(&rd, c.cfg, converted)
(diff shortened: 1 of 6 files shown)


c2thorn:
Looks like there's a panic caused by https://github.com/GoogleCloudPlatform/terraform-google-conversion/pull/709

melinath (changes requested):
LGTM as an implementation; would it be possible to add an integration test? I assume we would need to avoid using a module, but perhaps we could explicitly set up the underlying relationship that caused the crash?

Adding tests for resources is documented a bit here: https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/docs/add_new_resource.md#2-terraform-validator

I imagine it would be similar to the example_storage_bucket_iam_member tests.

melinath:
The tests are passing for both me & @c2thorn locally. This is because the results of `convert` are not fully deterministic.

We use [`map[string]Asset`](https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/converters/google/convert.go#L168) to store assets internally; Golang maps are not ordered. So when we pull out the list of assets, we order them by [Asset Name](https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/converters/google/convert.go#L342) - for example, `//storage.googleapis.com/placeholder-c2WD8F2q` (the last bit is generated by terraform-google-conversion).

In test runs, we then "normalize" assets for comparison, which replaces `placeholder-.*` with `placeholder-foobar` so that we can have passing tests even with randomly-generated asset names. This obscures the fact that the assets were first sorted by their random name (which means that the results are not deterministic even though they're sorted by name.)

melinath:
It's odd that it seems to pass consistently for me locally though... :-/

melinath:
Okay, so if I am just running the one test, it always gets the same values for the random placeholder strings. This is almost certainly because we use [`rand.Intn`](https://github.com/GoogleCloudPlatform/terraform-google-conversion/blob/a96c3de41339f92e9fb31b5ac62ec9b350484925/google/cai.go#L102) to generate the placeholder values, which relies on Golang's pseudo-random builtins, which always start with the [same randomness seed](https://golang.org/pkg/math/rand/#Seed).

So, the full test suite is probably failing because it's preceded by a bunch of `randIntn` calls, resulting in output that happens to be different than what we get just running that test. We're lucky we caught this now!

On the plus side, this means that if we can find a thread-safe way to reset the seed before each test, we could make all tests deterministic with regard to their generated strings. If that doesn't work we'll have to find some other solution.

c2thorn:
@melinath Moved the test failure to its own issue, skipped the test and rebased.

melinath (approved):
LGTM

PR #221, opened 2021-06-02
Add support for google_compute_network and google_compute_subnetwork

### LINKED ISSUE #20: Add support for google_compute_subnetwork
~- [ ] google_compute_network~ #211
- [ ] google_compute_subnetwork
- [x] google_compute_firewall

### LINKED ISSUE #211: Add support for google_compute_network
Hello Guys,

I am new to terraform-validator and I was running some tests with it.

I attempted the following:

 - Creating a default network with "google_compute_network"
 - Assigned a member a serviceAccountUser IAM role
 - Assigning an External IP to a compute instance

Of the 3 tasks, only the last one reported a violation and I keep on getting an error: "unsupported resource: google_compute_network" for the network and no violation was reported for the IAM configuration. 

image

image

The config validator scanner was able to report all 3 violations.

jday29:
Adds support for the `google_compute_network` (#20, #211) and `google_compute_subnetwork` (#20) resource types.

Note: The integration tests failed for me when running within the Docker container (`make build-docker`, `make run-docker`, `make test-integration`) because the `Dockerfile` is installing [Terraform `0.12.20`](https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/Dockerfile#L6), but one of the tests expects [Terraform `0.12.24`](https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/testdata/templates/example_project_update.tfstate#L3). I updated the version in the `Dockerfile` to install `0.12.24` to get around this issue. Not sure if that impacts any other tests/dependencies, or if we should jump to the latest Terraform `0.12.x` version (`0.12.31` as of Jun 6, 2021)?

It seems like `terraform-google-conversion` supports a large number of resources already, so I didn't update that dependency. Assuming we don't have to update it for the ones listed at https://github.com/GoogleCloudPlatform/terraform-google-conversion/tree/master/google, and these changes cover everything required for additional resource types, I may add others that I would like to validate against that aren't currently supported.

Resolved #20. Resolved #211.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM golang:1.14
 RUN apt-get update && apt-get -y install wget unzip
 
 WORKDIR /tmp
-RUN wget https://releases.hashicorp.com/terraform/0.12.20/terraform_0.12.20_linux_amd64.zip && unzip terraform_0.12.20_linux_amd64.zip -d /usr/local/bin
+RUN wget https://releases.hashicorp.com/terraform/0.12.24/terraform_0.12.24_linux_amd64.zip && unzip terraform_0.12.24_linux_amd64.zip -d /usr/local/bin
 
 
 ENV GO111MODULE=on
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ google_compute_forwarding_rule
 google_compute_global_forwarding_rule
 google_compute_firewall
 google_compute_instance
+google_compute_network
 google_container_cluster
 google_container_node_pool
 google_filestore_instance
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ google_compute_global_forwarding_rule
 google_compute_firewall
 google_compute_instance
 google_compute_network
+google_compute_subnetwork
 google_container_cluster
 google_container_node_pool
 google_filestore_instance
--- a/converters/google/mappers.go
+++ b/converters/google/mappers.go
@@ -56,6 +56,7 @@ func mappers() map[string][]mapper {
 		"google_compute_global_forwarding_rule": {{convert: converter.GetComputeGlobalForwardingRuleCaiObject}},
 		"google_compute_instance":               {{convert: converter.GetComputeInstanceCaiObject}},
 		"google_compute_network":                {{convert: converter.GetComputeNetworkCaiObject}},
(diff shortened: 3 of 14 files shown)


melinath (changes requested):
The CI/CD pipeline is running 12.24 so upgrading the docker container to match is great. Since we only use terraform to run tests, we're okay for now with not being on the latest version (unless someone reports an issue.) In any case, that would need to be a separate discussion from adding these resources. :-)

It looks like a bunch of integration tests are failing:

- TestCLI/v=0.12/tf=full_compute_firewall/offline=true/cmd=convert
- TestCLI/v=0.12/tf=full_compute_firewall/offline=false/cmd=convert
- TestCLI/v=0.12/tf=full_sql_database_instance/offline=false/cmd=convert
- TestCLI/v=0.12/tf=example_compute_firewall/offline=false/cmd=convert
- TestCLI/v=0.12/tf=full_sql_database_instance/offline=true/cmd=convert
- TestCLI/v=0.12/tf=example_compute_firewall/offline=true/cmd=convert

At a quick glance, the problem seems to be that those test cases also (have to?) include compute network resources; now that you've added support for those resources, they also need to be added to the expected output.

melinath (approved):

jday29:
> The CI/CD pipeline is running 12.24 so upgrading the docker container to match is great. Since we only use terraform to run tests, we're okay for now with not being on the latest version (unless someone reports an issue.) In any case, that would need to be a separate discussion from adding these resources. :-)
> 
> It looks like a bunch of integration tests are failing:
> 
> * TestCLI/v=0.12/tf=full_compute_firewall/offline=true/cmd=convert
> * TestCLI/v=0.12/tf=full_compute_firewall/offline=false/cmd=convert
> * TestCLI/v=0.12/tf=full_sql_database_instance/offline=false/cmd=convert
> * TestCLI/v=0.12/tf=example_compute_firewall/offline=false/cmd=convert
> * TestCLI/v=0.12/tf=full_sql_database_instance/offline=true/cmd=convert
> * TestCLI/v=0.12/tf=example_compute_firewall/offline=true/cmd=convert
> 
> At a quick glance, the problem seems to be that those test cases also (have to?) include compute network resources; now that you've added support for those resources, they also need to be added to the expected output.

Sorry about that, I forgot to run the integration tests against all cases; I only tested mine before merging.

melinath:
no worries! Thanks for the update :-)

PR #229, opened 2021-06-14
Wrote a release script

melinath:
This is aimed at making semver releases easier.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -10,8 +10,6 @@ fi
 
 version=$1
 
-echo $version | grep -Eq "^v[0-9]+\.[0-9]+\.[0-9]+$" && echo "match" || echo "no match"
-
 if ! echo $version | grep -Eq "^v[0-9]+\.[0-9]+\.[0-9]+$"
 then
 	echo "Invalid version: ${version}"
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [ $# -eq 0 ]
+then
+	echo "No version supplied. Run \`make release VERSION=X.X\`"
+	exit 1
+fi
+
+version=$1
+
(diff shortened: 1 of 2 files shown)


release.sh
```diff
@@ -0,0 +1,45 @@
...
+then
+	echo "No version supplied. Run \`make release VERSION=X.X\`"
+	exit 1
+fi
+
+version=$1
+
+echo $version | grep -Eq "^v[0-9]+\.[0-9]+\.[0-9]+$" && echo "match" || echo "no match"
+
+if ! echo $version | grep -Eq "^v[0-9]+\.[0-9]+\.[0-9]+$"
```
ScottSuarez, line 15: computed twice. both above and below. Recommend storing in variable

release.sh
```diff
@@ -0,0 +1,45 @@
...
+platforms="linux windows darwin"
+build_dir=./bin
+name=terraform-validator
+ldflags="-X github.com/GoogleCloudPlatform/terraform-validator/tfgcv.buildVersion=${version}"
+release_bucket=terraform-validator
+
+# Build release versions
+for platform in ${platforms}; do
+	echo "Building version ${version} for ${platform}..."
+	GO111MODULE=on GOOS=${platform} GOARCH=amd64 CGO_ENABLED=0 go build -ldflags "${ldflags}" -o "${build_dir}/${name}-${platform}-amd64" .
```
ScottSuarez, line 31: should we add support for apple silicon or other architectures?

release.sh
melinath: Whoops, the upper one was just for testing & I forgot to remove it. Thanks!

release.sh
melinath: Possibly? I don't know that we have a clear signal that we're missing particular platforms... But either way, I think that's outside the scope of this PR.

ScottSuarez (approved):

PR #231, opened 2021-06-21
new bucket IAM mappings

rkarger-google:
@melinath 

Added support for new bucket IAM mapping functions.


melinath (changes requested):
This needs to also update the terraform-google-conversion dependency - see https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/docs/add_new_resource.md#2-terraform-validator

melinath (approved):

PR #233, opened 2021-06-23
Moved list of supported resources into a separate doc

melinath:
Also added documentation for list-supported-resources. The idea here is that people will have a better overview of the documentation if the first thing isn't an ever-expanding list of resources.


slevenick (approved):
Is it possible to automate the generation of that list? I know from experience that it is super easy to forget to add new resources to lists like this!

melinath:
@slevenick possibly, but I'm not sure exactly how we'd do it or whether it's worth it... but I'm hoping that adding a disclaimer will help people recognize that it's going to potentially be out of sync.

PR #240, opened 2021-06-27
upgrade opa to v0.29.3

ericyz:


melinath:
It looks like this is a follow-up to https://github.com/forseti-security/config-validator/pull/166

melinath (approved):

Issue #246, opened 2021-07-02
Improve error message if --project isn't provided and a resource needs it


melinath:
Right now, if a resource is missing an expected `project` value and no default is set via the `--project` flag, we give an error like:

```
required field 'project' is not set
```

We should expand this error to tell the user how to resolve the issue - for example, let them know that they can resolve the issue by supplying `--project` to provide a default project.

krishnaindani:
@melinath Is this issue open to work? I would like to check if I can work on this.

melinath:
@krishnaindani yeah, you're definitely welcome to work on this!

krishnaindani:
Great. Thanks

krishnaindani:
@melinath I have opened the PR and updated the error message, please let me know if looks good or needs to be updated. Thank you.

krishnaindani:
@melinath Appreciate the opportunity to let me work on this.

PR #266, opened 2021-07-25
update error message when project is not passed

krishnaindani:
Resolved https://github.com/GoogleCloudPlatform/terraform-validator/issues/246

This updates the error message when the project is not passed.

### DIFF
--- a/converters/google/vendor_utils.go
+++ b/converters/google/vendor_utils.go
@@ -62,5 +62,5 @@ func getProjectFromSchema(projectSchemaField string, d converter.TerraformResour
 	if config.Project != "" {
 		return config.Project, nil
 	}
-	return "", fmt.Errorf("required field '%s' is not set", projectSchemaField)
+	return "", fmt.Errorf("required field '%s' is not set, you may use --project=my-project to provide default project to resolve the issue", projectSchemaField)
 }
--- a/converters/google/vendor_utils.go
+++ b/converters/google/vendor_utils.go
@@ -62,5 +62,5 @@ func getProjectFromSchema(projectSchemaField string, d converter.TerraformResour
 	if config.Project != "" {
 		return config.Project, nil
 	}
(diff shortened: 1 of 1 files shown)


krishnaindani:
@googlebot I signed it!

converters/google/vendor_utils.go
```diff
@@ -62,5 +62,5 @@ func getProjectFromSchema(projectSchemaField string, d converter.TerraformResour
 	if config.Project != "" {
 		return config.Project, nil
 	}
-	return "", fmt.Errorf("required field '%s' is not set", projectSchemaField)
+	return "", fmt.Errorf("required field '%s' is not set, you may use --project=my-project to provide default project to resolve the issue", projectSchemaField)
```
melinath, line 65: could this be: provide _a_ default project

melinath (changes requested):
LGTM - just a small change request

converters/google/vendor_utils.go
krishnaindani: so something simply like 
```suggestion
	return "", fmt.Errorf("required field '%s' is not set, provide a default project", projectSchemaField)
```

melinath: no, the version you had before was great, it was just missing the word "a" between `provide` and `default project`. So the full message would be:

```
required field '%s' is not set, you may use --project=my-project to provide a default project to resolve the issue
```

krishnaindani: oh okay, sorry I got that wrong. This is clear now, I will push the changes. Thanks

melinath: No worries! Thanks for making the update; I'm looking forward to incorporating this change!

krishnaindani: You're welcome. Updated the changes.

melinath (approved):
LGTM - thanks!

krishnaindani:
> LGTM - thanks!

Thank you.

Issue #268, opened 2021-07-27
KMS key ring & crypto key use incorrect domain in asset type and asset name


melinath:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave _+1_ or _me too_ comments; they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If the issue is assigned to a user, that user is claiming responsibility for the issue.

### Terraform Validator version

terraform-validator: v0.4.0

### Affected Resource(s)

* google_kms_crypto_key
* google_kms_key_ring

### Expected Behavior

According to https://cloud.google.com/asset-inventory/docs/supported-asset-types, these KMS resources should have the following asset types:

- cloudkms.googleapis.com/KeyRing
- cloudkms.googleapis.com/CryptoKey

The domain (cloudkms.googleapis.com) should also be used in the canonical asset name. (Note: it's also unclear whether the canonical asset name is correctly formatted but that is out of scope of this ticket.)

### Actual Behavior

The code currently uses the domain `kms.googleapis.com` instead: https://github.com/GoogleCloudPlatform/terraform-google-conversion/blob/9b13c676d981af90f533cd9526243355a4199f2a/google/kms_crypto_key.go#L24

### References

* This was discovered while reviewing https://github.com/GoogleCloudPlatform/magic-modules/pull/4997

ref: b/199763427

melinath:
@palani-ram-google-partner is working on a fix for this

palani-ram-google-partner:
@melinath I have worked on this and have a fix ready to go. please assign this to me, so i can make a PR.

melinath:
b/199763427

melinath:
I think the issue here is that Terraform Validator is using the [product name](https://github.com/GoogleCloudPlatform/magic-modules/blob/21d9f76cae689ed1efd19c2079d7f49e884cf94a/mmv1/templates/validator/resource_conversion.go.erb#L25) (kms) to generate the domain; we should maybe instead be using the [GA base_url](https://github.com/GoogleCloudPlatform/magic-modules/blob/21d9f76cae689ed1efd19c2079d7f49e884cf94a/mmv1/products/kms/api.yaml#L20).

We should fix this for IAM as well as for resources.

PR #277, opened 2021-08-01
Add kms iam validators

palani-ram-google-partner:
- Update Google Conversion version to use KMS IAM validators
- Add KMS IAM key ring IAM validator tests
- Add KMS IAM crypto key IAM validator tests


palani-ram-google-partner:
@melinath , Please take a look at it and let me how what makes the build fail. As a google partner, I do not have access to the cloudbuild ci failure details.

melinath (changes requested):
shared information - let me know if you have any other questions!

palani-ram-google-partner:
@melinath the tests are fixed and the ci pipeline passed. please review / merge this PR when you get a chance. Thanks.

palani-ram-google-partner:
@melinath rebased and ci passed.

test/cli_test.go
```diff
@@ -103,17 +109,29 @@ func TestCLI(t *testing.T) {
...
+		"TestCLI/v=0.12/tf=example_organization_iam_member/offline=false/cmd=validate/constraint=always_violate":       "skip because test runner doesn't have org permissions",
+		"TestCLI/v=0.12/tf=example_project_iam/offline=false/cmd=convert":                                              "example_project_iam is too complex to untangle merges with online data generically",
+		"TestCLI/v=0.12/tf=example_storage_bucket_iam_member_random_suffix/offline=false":                              "test produces inconsistent results based on randomized names - github.com/GoogleCloudPlatform/terraform-validator/issues/259",
+		"TestCLI/v=0.12/tf=example_storage_bucket_iam_member_random_suffix/offline=false/cmd=convert":                  "test produces inconsistent results based on randomized names - github.com/GoogleCloudPlatform/terraform-validator/issues/259",
+		"TestCLI/v=0.12/tf=example_kms_key_ring_iam_binding/offline=false/cmd=convert":                                 "skip because test keyring does not exist in project",
```
melinath, line 123: These tests shouldn't be skipped - the terraform file needs to be written a little differently instead. I'll explain in more detail in a comment on one of the terraform files.

testdata/templates/example_kms_key_ring_iam_member.tf
```diff
@@ -0,0 +1,34 @@
...
+    }
+  }
+}
+
+provider "google" {
+  {{if .Provider.credentials }}credentials = "{{.Provider.credentials}}"{{end}}
+}
+
+resource "google_kms_key_ring_iam_member" "key_ring" {
+  key_ring_id = "{{.Provider.project}}/global/keyring-example"
```
melinath, line 31: Instead of supplying a key_ring_id, this should use a reference to a key ring resource id. For example:

```
resource "google_kms_key_ring" "example_keyring" {
  name     = "keyring-example"
  location = "global"
  project  = "{{.Provider.project}}"
}
```

This is the way we expect users to interact with these resources, and will avoid the issue of the resource not existing.

melinath (changes requested):

melinath:
Hey Palani - This looks good for the most part. The one thing is that it would be great to get the online tests passing; it should be possible to do if you include the parent resource in the test terraform. I've given one example above.

testdata/templates/example_kms_key_ring_iam_member.tf
melinath: When I try this locally, I get the following error:

```
Error: converting tfplan to CAI assets: adding resource changes to converter: adding resource create or update fetching asset: Error parsing resource ID for : Invalid KeyRing id format, expecting `{projectId}/{locationId}/{keyRingName}` or `{locationId}/{keyRingName}.`
```

It looks like this is probably because there are a couple of lines missing from the kms_key_ring `Fetch*` code, which check for an empty parent field. (They were added to the template after you'd already started your branch, I think - sorry I didn't catch this sooner.)

Here are the corresponding lines for storage bucket iam: https://github.com/GoogleCloudPlatform/terraform-google-conversion/blob/a4c23f33b71b2fc62419b0e034e39fea654bb0c3/google/storage_bucket_iam.go#L77

You'll need to make a PR to add those lines to the terraform-google-conversion code for kms key ring iam & kms crypto key iam, then update the version of terraform-google-conversion used in this PR.

test/cli_test.go
palani-ram-google-partner: @melinath , this is fixed

testdata/templates/example_kms_key_ring_iam_member.tf
palani-ram-google-partner: @melinath , this is taken care. please review this PR.

palani-ram-google-partner:
@melinath , all changes are done. please review this PR.

melinath:
@palani-ram-google-partner merged https://github.com/GoogleCloudPlatform/magic-modules/pull/5074. When this PR is ready for review, you may be able to "re-request review" by clicking this button:

Screen Shot 2021-08-11 at 8 20 02 AM

melinath (changes requested):
It looks like `make test` is currently failing - there's a trailing hardcoded project name.

It looks like the integration tests are failing due to the asset name being incorrect - e.g. `"asset_type":"cloudkms.googleapis.com/KeyRing"` vs `"asset_type":"kms.googleapis.com/KeyRing"`.

The cloudkms version is correct. I feel like we have talked about how we're handling this for now but I forget the result of the conversation. Do you remember us talking about that? And do you remember where we had that conversation?

melinath:
ah, it looks like this issue is covered by https://github.com/GoogleCloudPlatform/terraform-validator/issues/268 - so it sounds like you have a fix prepared? We should merge that into TGC first.

melinath:
On reflection, I think it would be okay to merge this using `kms.googleapis.com` consistently, if that's possible, since there's a separate ticket to fix the mapping.

PR #278, opened 2021-08-02
Allow converting no-op resource change actions.

ankurdua:
- Added a new option "**convertUnchanged**" to Converter struct which allows converting unchanged/no-op resource changes and modified relevant methods for supporting it.
- Modified ReadPlannedAssets function to support this new option.

### DIFF
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -127,9 +127,9 @@ type RestoreDefault struct {
 }
 
 // NewConverter is a factory function for Converter.
-func NewConverter(ctx context.Context, ancestryManager ancestrymanager.AncestryManager, project string, offline bool) (*Converter, error) {
+func NewConverter(ctx context.Context, ancestryManager ancestrymanager.AncestryManager, project string, offline, convertUnchanged bool) (*Converter, error) {
 	cfg := &converter.Config{
-		Project:     project,
+		Project: project,
 	}
 	// Search for default credentials
 	cfg.Credentials = multiEnvSearch([]string{
@@ -149,12 +149,13 @@ func NewConverter(ctx context.Context, ancestryManager ancestrymanager.AncestryM
 	}
 
 	return &Converter{
-		schema:          provider.Provider(),
-		mapperFuncs:     converter.Mappers(),
-		offline:         offline,
-		cfg:             cfg,
-		ancestryManager: ancestryManager,
-		assets:          make(map[string]Asset),
+		schema:           provider.Provider(),
+		mapperFuncs:      converter.Mappers(),
+		offline:          offline,
+		cfg:              cfg,
+		ancestryManager:  ancestryManager,
+		assets:           make(map[string]Asset),
+		convertUnchanged: convertUnchanged,
 	}, nil
 }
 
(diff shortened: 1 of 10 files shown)


ankurdua:
Hi @melinath, PTAL and let me know your thoughts on this. Thanks.

converters/google/convert.go
```diff
@@ -291,10 +295,10 @@ func (c *Converter) addDelete(rc *tfjson.ResourceChange) error {
 	return nil
 }
 
-// For create/update, we need to handle both the case of no merging,
+// For create/update/no-op, we need to handle both the case of no merging,
 // and the case of merging. If merging, we expect both fetch and mergeCreateUpdate
 // to be present.
-func (c *Converter) addCreateOrUpdate(rc *tfjson.ResourceChange) error {
+func (c *Converter) addResourceChange(rc *tfjson.ResourceChange) error {
```
melinath, line 301: Now that I'm seeing the implementation, I think maybe this should be `addCreateOrUpdateOrNoop` after all; it's longer but perhaps more clear (and matches `createOrUpdateOrNoops` above). But I think it could be fine either way (and it would be easy to change later since it's private). What do you think?

converters/google/convert.go
```diff
@@ -240,7 +244,7 @@ func (c *Converter) AddResourceChanges(changes []*tfjson.ResourceChange) error {
 // both fetch and mergeDelete. Supporting just one doesn't
 // make sense, and supporting neither means that the deletion
 // can just happen without needing to be merged.
-func (c *Converter) addDelete(rc *tfjson.ResourceChange) error {
+func (c *Converter) addDeleteResourceChange(rc *tfjson.ResourceChange) error {
```
melinath, line 247: If the other method does switch to `addCreateOrUpdateOrNoop`, this should go back to `addDelete`

melinath (changes requested):
This looks great! I just have a couple of questions & suggestions.

converters/google/convert_test.go
```diff
@@ -212,52 +213,93 @@ func TestAddResourceChanges_noopIgnored(t *testing.T) {
...
-				},
-				"physical_block_size_bytes": 4096,
-			},
-			After: nil,
+	cases := []struct {
+		name             string
+		convertUnchanged bool
+	}{
+		{
+			name:             "Delete when convertUnchanged is unset",
```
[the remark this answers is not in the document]
ankurdua: Fixed.

converters/google/convert_test.go
```diff
@@ -212,52 +213,93 @@ func TestAddResourceChanges_noopIgnored(t *testing.T) {
...
+			convertUnchanged: false,
+		},
+		{
+			name:             "Delete when convertUnchanged is set",
```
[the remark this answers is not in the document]
ankurdua: Fixed.

converters/google/convert_test.go
```diff
@@ -212,52 +213,93 @@ func TestAddResourceChanges_noopIgnored(t *testing.T) {
...
-		name    string
-		actions tfjson.Actions
+		name             string
+		actions          tfjson.Actions
+		convertUnchanged bool
 	}{
 		{
-			name:    "Create",
-			actions: tfjson.Actions{"create"},
+			name:             "Create when convertUnchanged is unset",
```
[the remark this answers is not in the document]
ankurdua: Fixed.

converters/google/convert.go
ankurdua: `addCreateOrUpdateOrNoop` is definitely more clear. Renamed.

converters/google/convert.go
ankurdua: Renamed back to `addDelete`.

melinath (approved):
LGTM

Issue #279, opened 2021-08-03
GOOGLE_OAUTH_ACCESS_TOKEN is not being properly utilized


mscifo:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave _+1_ or _me too_ comments; they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If the issue is assigned to a user, that user is claiming responsibility for the issue.

### Terraform Validator version

terraform-validator: v0.4.0

### Affected Resource(s)

* google

### Terraform Plan JSON

N/A

### Debug Output

```
Error: converting tfplan to CAI assets: constructing resource manager client: constructing resource manager client: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
Usage:
  terraform-validator validate <tfplan> [flags]
```

### Expected Behavior

That terraform-validator will properly use the environment variable `GOOGLE_OAUTH_ACCESS_TOKEN` for google credentials.

### Actual Behavior

The `GOOGLE_OAUTH_ACCESS_TOKEN` variable appears to be ignored.

### Steps to Reproduce

1. `rm -rf ~/.config/gcloud/*`   # WARNING: don't actually do this unless you want to wipe all your gcloud credentials
2. `GOOGLE_OAUTH_ACCESS_TOKEN=<REDACTED> terraform-validator validate plan.json --policy-path=./policy_library --project <REDACTED>`

### Important Factoids

None

melinath:
@ScottSuarez It looks like the code is properly extracting the access token from the environment; I'm not immediately sure what the issue here is. In case it helps with investigation, the code we've got is based on what the terraform provider does.

ScottSuarez:
hi @melinath , found fix.. tpg/b token initialization is done before load and validate. I aligned the token/credential searching in pr. Should fix

ScottSuarez:
scratch that, new possibility, will update in time

PR #281, opened 2021-08-04
Synced provider credential initialization to be in line with tpg/b

ScottSuarez:
fixes https://github.com/GoogleCloudPlatform/terraform-validator/issues/279

credentials/token were not being initialized from before calling `LoadAndValidate`. 

see https://github.com/hashicorp/terraform-provider-google/blob/master/google/provider.go#L1270-L1279 for tpg/b implementation.

### DIFF
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -127,11 +127,24 @@ type RestoreDefault struct {
 }
 
 // NewConverter is a factory function for Converter.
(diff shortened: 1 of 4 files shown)


melinath (changes requested):
This seems to be based off an old commit - could you rebase this off of `main`? (The diff is currently a little hard to read & there are merge conflicts.)

Also, is there possibly a way to write a test for this?

PR #287, opened 2021-08-11
Bundled third party notices and source code with executable for release

### LINKED ISSUE #272: Bundle dependency licenses and source code with binaries
We want to bundle dependency licenses and source code (as required) with released binaries. This requires switching to a tarball for distribution instead of just a binary.

As part of this change, we're going to switch to the following naming convention instead of the current one:

```
terraform-validator_[PLATFORM]_[ARCH]-[VERSION].tar.gz
```

...where:

- PLATFORM is linux, darwin, or windows
- ARCH is `amd64` or `arm`
- VERSION is the semantic version without a `v` prefix, e.g. 0.1.0

The tarball will be flat and contain the following files:

- the binary (as `terraform-validator`)
- the source code & licenses (as a zip file)

melinath:
Resolved #272

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -10,20 +10,35 @@ fi
 
 version=$1
 
-if ! echo $version | grep -Eq "^v[0-9]+\.[0-9]+\.[0-9]+$"
+if ! echo $version | grep -Eq "^[0-9]+\.[0-9]+\.[0-9]+$"
 then
 	echo "Invalid version: ${version}"
-	echo "Please specify a semantic version prefixed with a v (e.g. v0.0.0)."
+	echo "Please specify a semantic version with no prefix (e.g. X.X.X)."
 	exit 1
 fi
+
+release_dir="./release/${version}"
+rm -rf $release_dir
+mkdir -p $release_dir
+
+echo "Go version: $(go version)"
+go get github.com/google/go-licenses
+echo "Downloading licenses and source code to bundle..."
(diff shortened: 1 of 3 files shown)


release.sh
```diff
@@ -10,20 +10,35 @@ fi
...
+
+echo "Go version: $(go version)"
+go get github.com/google/go-licenses
+echo "Downloading licenses and source code to bundle..."
+# Ignore errors until https://github.com/google/go-licenses/pull/77 is merged
+set +e
+go-licenses save "github.com/GoogleCloudPlatform/terraform-validator" --save_path="./${release_dir}/THIRD_PARTY_NOTICES"
+set -e
+echo "Zipping licenses and source code..."
+pushd "${release_dir}" > /dev/null
```
ScottSuarez, line 32: `> /dev/null`
What does this do?

ScottSuarez (approved):
question on [ticket](https://github.com/GoogleCloudPlatform/terraform-validator/issues/272#issuecomment-897129971)

release.sh
melinath: TLDR; normally pushd outputs the current stack of directories, but that output is useless and distracting in this context. `> /dev/null` suppresses that output.

The way it works is that `>` directs the stdout of the command to a file; `/dev/null` is a special file descriptor that silently drops everything written to it.

Here's more information about this: https://unix.stackexchange.com/questions/119648/redirecting-to-dev-null

PR #303, opened 2021-09-02
Ignore asset fetch with a warning if user does not have read permissions

### LINKED ISSUE #290: Permissions for validating *_iam resources
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave _+1_ or _me too_ comments; they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If the issue is assigned to a user, that user is claiming responsibility for the issue.

In `v2021-03-22` no additional IAM permissions were needed for running tfv on resources which included IAM policies.

However starting with `v2021-04-07` an error is thrown 
```
 Error: converting tfplan to CAI assets: adding resource changes to converter: adding resource create/update/no-op fetching asset: Error retrieving IAM policy for organization "ORG_ID": googleapi: Error 403: The caller does not have permission, forbidden
 ```

It looks like additional IAM permissions maybe needed, I was able to get it running with `Security Reviewer`. I was wondering if we can add some additional notes to [auth docs](https://github.com/GoogleCloudPlatform/terraform-validator/blob/main/docs/user_guide.md#auth)

### Terraform Validator version

terraform-validator: v2021-04-07+

### Affected Resource(s)

* `google_*_iam_*`

### Terraform Plan JSON

go/tfv-debug

### PR DESCRIPTION
**Pull request by melinath**
Resolved #290. This can be tested manually but it's a bit tricky. You have to:

1. create a service account that doesn't have any permissions
2. Set up `GOOGLE_IMPERSONATE_SERVICE_ACCOUNT=<that account>`
3. Generate a plan from a terraform file like:
   ```
   resource "google_project_iam_member" "viewer-a" {
     project = "my-project"
     role    = "roles/viewer"
     member  = "user:example-a@google.com"
   }
   ```
4. build terraform-validator and use it to convert the plan. You should get a warning due to lacking read permissions during "asset fetch", and then fail during getting resource ancestry (rather than failing due to lacking read permissions during asset fetch).

Unfortunately there's not currently a great way to add automated tests for this right now.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -37,6 +37,10 @@ GOOGLE_APPLICATION_CREDENTIALS=~/.config/gcloud/application_default_credentials.
 
 You can specify a [service account to impersonate](https://cloud.google.com/iam/docs/impersonating-service-accounts) for all Google API calls with the `GOOGLE_IMPERSONATE_SERVICE_ACCOUNT` environment variable.
 
+### Validating compliance IAM member & binding resources
+
+Validating compliance for an IAM policy requires the full policy. However, the Terraform Provider's IAM member & binding resources only manage a slice of an IAM resource. In order to work around this, Terraform Validator fetches the full IAM resource from GCP and merges it with the Terraform resources prior to validation. This means that users who are validating IAM member and binding resources need to have permissions to read IAM policies - for example, via the Security Reviewer role.
+
 ## `terraform-validator validate`
 
 This command allows you to validate your terraform plan JSON against a specific policy library.
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -37,9 +37,14 @@ GOOGLE_APPLICATION_CREDENTIALS=~/.config/gcloud/application_default_credentials.
 
 You can specify a [service account to impersonate](https://cloud.google.com/iam/docs/impersonating-service-accounts) for all Google API calls with the `GOOGLE_IMPERSONATE_SERVICE_ACCOUNT` environment variable.
 
-### Validating compliance IAM member & binding resources
+### Required permissions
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by ScottSuarez** [comment]
Should we be ignoring this ?

Seems appropriate to throw an error if we can retrieve stuff? 
Maybe it shouldn't crash but something should definitely be printed at the end of application cycle and application shouldn't exit with 0

**Comment by ScottSuarez** [comment]
`and then fail during getting resource ancestry` maybe we can sync on this so I understand the system design

**Comment by melinath** [comment]
Per discussion:

- This is a little different than the case where a parent resource ID is missing, since it's a user configuration issue rather than an inherent thing about how terraform plans work.
- We believe the default behavior should be to pull down the IAM resources, since it is the only way to get accurate results for validating them.
- So far, nobody is asking for a way to disable this behavior; if that ever becomes an ask, we can deal with it at that time.

@ScottSuarez I've switched this to a docs-only change.

File: docs/user_guide.md [hunk 1]
```diff
@@ -37,6 +37,10 @@ GOOGLE_APPLICATION_CREDENTIALS=~/.config/gcloud/application_default_credentials.
 
 You can specify a [service account to impersonate](https://cloud.google.com/iam/docs/impersonating-service-accounts) for all Google API calls with the `GOOGLE_IMPERSONATE_SERVICE_ACCOUNT` environment variable.
 
+### Validating compliance IAM member & binding resources
```
ScottSuarez, line 40:
```suggestion
### Permission Requirements

The service account performing the validation must have the following permissions to avoid runtime issues fetching existing resources.

#### Security Reviewer role
```

ScottSuarez (approved):
Might help to tell the user what to do upfront and reasoning later if they need it.

ScottSuarez (approved):
lgtm, it might make sense to give users more broad roles (as a suggestion) that would encompass these in the future if this list becomes too long (if we expect this list to continue to grow). 

This works though!

Issue #255, opened 2021-07-16
Output errors & warnings as JSON


melinath:
In order to make it easier to parse error / warning data, we should start outputting errors as JSON using a schema like this:

```
{
    "version": "vX.X.X",
    "timestamp": "RFC 3339-encoded timestamp",
    "error_details": {
        "error": "error type",
        "context": "additional error context (optional)"
    }
}
```

PR #304, opened 2021-09-13
Basic implementation of structured logging

melinath:
Implemented the first steps of https://github.com/GoogleCloudPlatform/terraform-validator/issues/255. Essentially:

- Added in zap loggers to output logs in a structured way.
  - Use the zap loggers in all cases for errors, but default to Console logging
  - Continue defaulting to fmt.Printf for command output.
  - If the user passes a specific env var, switch to structured logging for errors and command output.
- Redirected stdlib logging to zap to capture it in a structured way as well
- Made validate / convert option structs store additional functions that we can set to mocks, so we can write unit tests for the commands

Remaining work after this PR will be to pipe the loggers through to the convert/validate internals and remove remaining usages of fmt.Printf, log, and glog.

As part of this work, I also turned on SilenceUsage and SilenceErrors at the root command, since that output is not generally useful and can't be made structured (since it's printed directly to stdout.) We instead log any returned errors in the main Execute func.


cmd/convert_test.go
```diff
@@ -0,0 +1,92 @@
...
+		readPlannedAssets:    MockReadPlannedAssets,
+	}
+
+	err := o.run("/path/to/plan")
+	a.Nil(err)
+
+	errorLogs := errorObservedLogs.AllUntimed()
+	outputLogs := outputObservedLogs.AllUntimed()
+
+	// On a successful legacy run, we don't output anything via loggers.
```
slevenick, line 89: Does logging output need to be gated by this flag? Does it impact functionality in some way?

slevenick (approved):

cmd/convert_test.go
melinath: My understanding is that users might be parsing this output; although the structured output might be more useful, it's in a different format, so switching it would be a breaking change for folks.

PR #306, opened 2021-09-16
Piped zap loggers throughout the conversion process

melinath:
With this change, all of our error / status logging should use the same zap logger, allowing us to enable structured logging for everything. This is a follow-up for https://github.com/GoogleCloudPlatform/terraform-validator/pull/304

### DIFF
--- a/ancestrymanager/ancestrymanager.go
+++ b/ancestrymanager/ancestrymanager.go
@@ -3,12 +3,12 @@ package ancestrymanager
 
 import (
 	"fmt"
-	"log"
 	"strings"
 
 	"google.golang.org/api/cloudresourcemanager/v1"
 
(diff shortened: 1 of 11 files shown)


c2thorn (approved):
Very minor/possibly not worthwhile opinion: `logger` instead of `errorLogger`? Or something else generic to denote it does more than actual errors.

melinath:
FWIW it's errorLogger in opposition to the outputLogger because it outputs to stderr (and the structured logging schema we're using treats everything that's going to stderr as an "error"). But outputLogger only gets used in the cmd package, so maybe it just ends up being confusing? For now I think I'll merge this as-is for consistency but we can re-evaluate later. It will at least be a searchable term.

Issue #309, opened 2021-09-22
Support for new resource "ServicePerimeter"


ankurdua:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave _+1_ or _me too_ comments; they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If the issue is assigned to a user, that user is claiming responsibility for the issue.

### Description

Add support for  "accesscontextmanager.googleapis.com/ServicePerimeter"  resource.

### New or Affected Resource(s)

* google_access_context_manager_service_perimeter

### Terraform-google-conversion support?

- [google_access_context_manager_service_perimeter is in terraform-google-conversion](https://github.com/GoogleCloudPlatform/terraform-google-conversion/blob/master/google/accesscontextmanager_service_perimeter.go)

PR #311, opened 2021-09-22
Add tests for google_access_context_manager_service_perimeter resource.

ankurdua:
https://github.com/GoogleCloudPlatform/terraform-validator/issues/309#issue-1004492601

- Added tests for new resource google_access_context_manager_service_perimeter.
- Updated terraform-google-conversion version.
- Updated default terraform google provider version used in tests.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/GoogleCloudPlatform/terraform-validator
 
 require (
-	github.com/GoogleCloudPlatform/terraform-google-conversion v0.0.0-20210917180804-7d769d84f6a1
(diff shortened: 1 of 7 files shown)


melinath:
It looks like the build process uses the image developer-tools built from here: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/tree/master/infra/build

I'll need to update that to use an image that supports 12.31 and then re-run this PR.

melinath:
There are no versions of that image that support TF 12.31 - this will take a little more effort to fix.

melinath (approved):
LGTM

PR #313, opened 2021-09-22
Run tests with multiple versions of terraform binary

### LINKED ISSUE #275: Run tests on multiple terraform versions
This would require building our own docker containers; however, it would allow us to have better control over our test environments and we could easily run integration tests against multiple versions of terraform.

melinath:
Also moved cloudbuild yaml to single location to clarify their relationship.

The primary reason for this change is to fix the ci/cd pipeline - but as long as I'm mucking with it, I can also close #275


melinath:
Note: the 0.14+ tests are failing because they require us to "finish" the upgrade to 0.13 first. That's a longer-term issue; I've removed those checks for now and will open a separate issue to look into finding a workaround. For now, we're at least supporting 2 versions.

rileykarson (approved):

PR #321, opened 2021-09-28
Switched to verbosity flag

melinath:
This replaces the boolean --verbose flag. Also, it turns out that the flag values actually get extracted from the command line during the Execute step, so the value of `--verbose` wasn't actually being used to initialize the loggers. I moved the logger initialization into PersistentPreRun to resolve this issue, since that [runs first](https://github.com/spf13/cobra/blob/4fd30b69ee2b62cf3bbecf0a423f8a1ee47f5f24/command.go#L98).

Note: About 1/3 of this PR is whitespace changes - reviewing with whitespace changes hidden might be easier.

### DIFF
--- a/cmd/convert.go
+++ b/cmd/convert.go
@@ -41,21 +41,17 @@ Example:
 `
 
 type convertOptions struct {
-	project              string
-	ancestry             string
-	offline              bool
-	errorLogger          *zap.Logger
-	outputLogger         *zap.Logger
-	useStructuredLogging bool
-	readPlannedAssets    tfgcv.ReadPlannedAssetsFunc
+	project           string
+	ancestry          string
+	offline           bool
+	rootOptions       *rootOptions
+	readPlannedAssets tfgcv.ReadPlannedAssetsFunc
 }
 
-func newConvertCmd(errorLogger, outputLogger *zap.Logger, useStructuredLogging bool) *cobra.Command {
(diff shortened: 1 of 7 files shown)


cmd/convert.go
```diff
@@ -41,21 +41,17 @@ Example:
...
-	ancestry             string
-	offline              bool
-	errorLogger          *zap.Logger
-	outputLogger         *zap.Logger
-	useStructuredLogging bool
-	readPlannedAssets    tfgcv.ReadPlannedAssetsFunc
+	project           string
+	ancestry          string
+	offline           bool
+	rootOptions       *rootOptions
```
rileykarson, line 47: Given any thought to https://golang.org/doc/effective_go#embedding? Seems like it fits what you're doing here pretty well.

rileykarson (approved):

cmd/convert.go
melinath: I did consider it, but I opted not to. Basically my thought process was that the child options are _using_ rootOptions rather than _extending_ it, if that makes sense? LMK if you have strong feelings either way; I'm happy to change it.

rileykarson: No strong feelings!

melinath: Okay, then I will merge as-is.

PR #328, opened 2021-10-10
KMS Crypto Key and Key Ring IAM Tests Added

palani-ram-google-partner:
KMS Crypto Key and Key Ring IAM Tests Added


palani-ram-google-partner:
@melinath Here is the TFV PR corresponding to MM PR https://github.com/GoogleCloudPlatform/magic-modules/pull/5293

All tests passed for TFV in my local. However something failed in cloud-build. Please take a look at it and let me know what fails?

melinath:
The previous issue was resolved with https://github.com/GoogleCloudPlatform/terraform-validator/pull/329.

This PR will need to be updated a bit due to https://github.com/GoogleCloudPlatform/terraform-validator/pull/331. Basically, magic-modules now generates the resources directly into TFV instead of needing to use TGC as a bridge. That means you can:

1. Remove the TGC/TPG updates from go.sum and go.mod
2. Run `make validator OUTPUT_PATH="/path/to/your/terraform-validator"` from mmv1 while on the upstream PR branch
3. Include the generated changes directly in this PR, which will let the unit & integration tests run correctly.

PR #360, opened 2021-11-01
Add Tests for Bigquery dataset IAM 

palani-ram-google-partner:
Add Tests for Bigquery dataset IAM


palani-ram-google-partner:
@melinath Please review this PR

Here is the Corresponding MM PR for this TFV
https://github.com/GoogleCloudPlatform/magic-modules/pull/5408

melinath (changes requested):
It looks like the converted asset files are missing the iam data. Could you add that, and also include the resources that are generated from the magic modules PR?

palani-ram-google-partner:
@melinath  Added the missing iam data in Converted asset files . Included the resources that are generated from the magic modules PR. Please review it.

melinath (changes requested):
iam_bigquery_dataset.go needs to be the same as the version in mmv1 - I'm not sure if that's what is causing the integration test failures but it might be worth getting in sync first just in case.

palani-ram-google-partner:
@melinath Fixed the iam_bigquery_dataset.go . Test is working fine. Please review it.

melinath (approved):
LGTM

PR #361, opened 2021-11-02
Merge offline and online ancestry manager

iyabchen:
- The merge changes the New function to take map[string][string] for project-ancestry pairs. 
- For GetAncestryWithResource, Online manager used to get from cache/API first, then get from resource. Offline manager used to get from resource first, then get from cache. Since there is GetAncestry function to fetch from cache/API, after the merge, GetAncestryWithResource will get from resource first, if it cannot find any, run GetAncestry. 
- For store() regarding cache, relax the condition to allow ancestry to be empty string, in order to pass test.


melinath:
It looks like the failure in `TestReadPlannedAssetsCoverage/full_sql_database_instance` is due to the following diff in the two lists of resources:

```diff
1c1
< listA:
---
> listB:
33c33
<     (string) (len=8) "settings": (map[string]interface {}) (len=16) {
---
>     (string) (len=8) "settings": (map[string]interface {}) (len=13) {
35,38d34
<      (string) (len=25) "authorizedGaeApplications": ([]interface {}) (len=2) {
<       (string) (len=32) "test-authorized_gae_application1",
<       (string) (len=32) "test-authorized_gae_application2"
<      },
46d41
<      (string) (len=27) "crashSafeReplicationEnabled": (bool) true,
85d79
<      (string) (len=15) "replicationType": (string) (len=21) "test-replication_type",
128,130c122
< }
< 
< 
---
> }
\ No newline at end of file
```

I'm not sure offhand why that's happening; if you want to investigate, let me know. I'm also happy to dig a little deeper.

ancestrymanager/ancestrymanager.go
```diff
@@ -126,52 +131,26 @@ func (m *onlineAncestryManager) GetAncestry(project string) (string, error) {
...
-}
-
-// GetAncestry returns the ancestry for the project. It returns an error if
-// the project does not equal to the one provided during initialization.
-func (m *offlineAncestryManager) GetAncestry(project string) (string, error) {
-	if project != m.project {
-		return "", fmt.Errorf("cannot fetch ancestry in offline mode")
+func (m *manager) store(project, ancestry string) {
+	if project != "" {
+		if m.offline {
```
melinath, line 136: We shouldn't need to have a separate check for offline vs online mode here because ancestry is required to be non-empty in offline mode.  https://github.com/GoogleCloudPlatform/terraform-validator/blob/21cb55867a0d21110e73c5ff8360b54f68daf48e/cmd/validate.go#L95

(If having a separate check for offline is required to get tests to pass the tests are probably wrong.)

melinath (changes requested):
LGTM apart from one change!

ancestrymanager/ancestrymanager.go
iyabchen: Restored this to the original version of online ancestry manager's store method. 
Have to change convert_test.go's newTestConverter() to fill in a non empty string for ancestry to make unit test work.

melinath (approved):
lgtm

PR #363, opened 2021-11-03
Fixed GetOk / GetOkExists behavior

### LINKED ISSUE #113: Error: converting tfplan to CAI assets: .. Error resolving image name ''"
**What do I see?**
When running a policy validate on our terraform, we get the following error.

`Error: converting tfplan to CAI assets: converting resource google_compute_disk: converting asset: Error resolving image name '': Error checking if image exists: json: invalid use of ,string struct tag, trying to unmarshal "projects/acmck-02-sandbox/global/images/" into uint64`

**What do I expect**
The policy validation  to output a violation(s) or report "No Violations".

**Reproduce**
Attach a disk to GCE instance with no image defined

```
resource "google_compute_disk" "bastion-disk" {
  name = "${var.name}-disk"
  type = "pd-ssd"
  zone = "us-central1-a"
  size = 100
}
```

and attach inside the instance resource:

```
    attached_disk {
      source      = google_compute_disk.bastion-disk.self_link
  }
```

**versions**

```
bash-5.0# terraform version
Terraform v0.12.16
+ provider.google v2.17.0
+ provider.google-beta v2.20.0
+ provider.kubernetes v1.10.0
+ provider.null v2.1.2
+ provider.random v2.2.1

Your version of Terraform is out of date! The latest version
is 0.12.17. You can update by downloading from https://www.terraform.io/downloads.html

bash-5.0# terraform-validator version
Build version: 2019-11-07
```
Happens on alpine or debain base image based containers.

```
bash-5.0# uname -a
Linux 9a1661522e53 4.9.184-linuxkit #1 SMP Tue Jul 2 22:58:16 UTC 2019 x86_64 Linux
```

melinath:
Resolved #113. Made FakeResourceData behave more consistently with Terraform core in terms of the values it produces. This ensures that expanders written with Terraform [ResourceData](https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/resource_data.go) structs in mind behave properly in TFV as well.

Some test data had to be updated due to the code now properly excluding empty data.

I'm sorry for the long PR; most of it is new tests.

### DIFF
--- a/converters/google/fake_resource_data.go
+++ b/converters/google/fake_resource_data.go
@@ -19,12 +19,16 @@ package google
 
 import (
 	"fmt"
+	"reflect"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
+
+// Must be set to the same value as the internal typeObject const
+const typeObject schema.ValueType = 8
 
 // Compare to https://github.com/hashicorp/terraform-plugin-sdk/blob/97b4465/helper/schema/resource_data.go#L15
 type FakeResourceData struct {
@@ -52,7 +56,7 @@ func (d *FakeResourceData) Get(name string) interface{} {
 // Get reads a single field by key and returns a boolean indicating
 // whether the field exists.
 func (d *FakeResourceData) GetOk(name string) (interface{}, bool) {
-	res, err := d.reader.ReadField(strings.Split(name, "."))
+	r, err := d.reader.ReadField(strings.Split(name, "."))
 	if err != nil {
 		return nil, false
 	}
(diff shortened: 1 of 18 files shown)


converters/google/fake_resource_data.go
```diff
@@ -43,6 +58,36 @@ func (d *FakeResourceData) Id() string {
...
+		s = schemaPath[len(schemaPath)-1]
+	}
+	if r.Value == nil && s != nil {
+		r.Value = r.ValueOrZero(s)
+	}
+
+	return getResult{
+		Value: r.Value,
+		Computed: r.Computed,
+		Exists: r.Exists,
```
ScottSuarez, line 86: will this ever be false? I don't really understand how this reader is working

melinath: r is the result from reading the field - Exists would be false if the field isn't set in the underlying data.

ScottSuarez (approved):

slevenick (approved):
Seems reasonable to me! Good to have this better represent normal Terraform functionality

PR #375, opened 2021-11-10
Added test files for the resource Compute attached disk

palani-ram-google-partner:
Added test files for the resource Compute attached disk


palani-ram-google-partner:
@melinath Here's the corresponding MM PR
https://github.com/GoogleCloudPlatform/magic-modules/pull/5441

palani-ram-google-partner:
@melinath Sorry it's my misunderstanding. I see that google compute resource is already added by someone else, which is exactly the resource I needed to validate. It shows up as part of the related Compute Disk CAI asset. This PR is not needed so I am closing it.

PR #379, opened 2021-11-15
Add tests for google_monitoring_notification_channel resource and add it to supported resources.

ankurdua:


ankurdua:
magic-modules PR:- https://github.com/GoogleCloudPlatform/magic-modules/pull/5453

ScottSuarez:
it seems this resource isn't a supported type in the CAI asset inventory 
https://cloud.google.com/asset-inventory/docs/supported-asset-types

From my understanding we  can't support this resource... Is that correct @melinath

melinath:
Hmm... it looks like it is a GA product so it should have CAI support - I'm not sure why it's not listed on that page, though. Maybe it's an oversight?

ankurdua:
As of today CAI does not support all GA GCP products. [Here](https://github.com/GoogleCloudPlatform/terraform-validator/blob/main/converters/google/resources/datalossprevention_job_trigger.go) is another one. But that is not a technical requirement for a resource to be added to validator right? Isn't only support in provider is sufficient?

melinath:
Terraform Validator is only intended to support assets that have a canonical CAI representation, in order to ensure that it's compatible with products that directly read from CAI. If we support resources that are not in CAI, and CAI support is added later, it could cause problems for related products.

We are currently generating code for datalossprevention_job_trigger but it's not actually used - the resource isn't actually supported. It's possible some resources have slipped through but if so that was unintentional. I'll check whether there's another place we should be looking for a canonical list of CAI asset types.

melinath:
Looking around, I feel fairly confident that this resource isn't supported by CAI - so I'll need to think about how we want to move forward here.

melinath:
After talking this through internally, we're currently not comfortable adding support for converting resources to CAI if they don't actually exist as CAI resources. Although it may offer some benefit to users, it would likely result in incompatibility with server-side checks that pull use CAI data directly.

PR #383, opened 2021-11-17
ci changes to run go-licenses before check-in

ScottSuarez:


melinath (changes requested):
It would be nice to run this as a separate cloudbuild trigger from cloudbuild-tests-unit - that way, they can fail independently and be clearly displayed in the github UI

ScottSuarez:
@melinath  presumably I'll have to merge this before I can set up such a task?

.ci/cloudbuild-tests-go-licenses.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+# See cloud-foundation-cicd project
+
+timeout: 1800s
+steps:
+- id: test-go-licenses
+  name: 'gcr.io/graphite-docker-images/terraform-validator-builder'
+  entrypoint: /bin/bash
+  args: ['-c', 'make test-go-licenses']
+  waitFor: ['build']
```
melinath, line 23: I think you will need to remove this line

```suggestion
```

Makefile
```diff
@@ -13,6 +13,11 @@ test-integration:
 	terraform --version
 	go test -v -run=CLI ./test
 
+test-go-licenses:
+	cd .. && go version && go get github.com/google/go-licenses
```
melinath, line 17: Does using a `cd ..` here mean that the following line will run in the parent directory or will the working directory reset?

melinath (approved):
You could set up the trigger first - it's just that all in-flight PRs will fail to run the check because the file doesn't exist yet. But that will happen regardless of the order. As long as you confirm that the cloudbuild trigger works right I think it should be fine.

A couple small things but otherwise LGTM.

Makefile
ScottSuarez: Each line within make is sandboxed. So it won't affect the subsequent line.

https://stackoverflow.com/questions/1789594/how-do-i-write-the-cd-command-in-a-makefile

melinath (approved):

PR #385, opened 2021-11-17
Upgrade 4.2.0

slevenick:
go.mod finds it as 3.0.0, which isn't quite right, but I'm not sure if that's important or how to change it. Maybe a tag?


melinath:
Oh, right, we're on v4 now - you would need to update the changes for https://github.com/hashicorp/terraform-provider-google/tree/release-4.2.0-v3 to use v4 instead of v3 and then I think it should work? But if that doesn't do it then this seems fine too, since the commit is still correct.

Issue #393, opened 2021-11-19
Error regarding: adding resource create/update/no-op fetching asset: lacking read permissions


ranjitk-burwood:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave _+1_ or _me too_ comments; they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If the issue is assigned to a user, that user is claiming responsibility for the issue.

### Terraform Validator version

terraform-validator: v0.9.1

### Affected Resource(s)

Unsure of which resource is affected, if any.

### Terraform Plan JSON

I can provide this if it is needed to further troubleshoot.

### Debug Output

https://gist.github.com/ranjitk-burwood/c8ae50b253e0e2443554cf7a2a5a212f

### Expected Behavior

I expected terraform-validator to validate a .json file.

### Actual Behavior

I am receiving an error about insufficient read permissions. My Cloud Build service account has Project IAM Admin at the Organization level and should have the necessary IAM permissions for resourcemanager.projects.get and resourcemanager.project.getIamPolicy.

1. `terraform plan -out=tfplan.tfplan`
2. `terraform show -json tfplan.tfplan > tfplan.json`
3. `terraform-validator validate tfplan.json --policy-path=<POLICY_PATH>`

### Important Factoids

This is being run in a Cloud Build pipeline that only runs up to a `terraform plan`. There is no current infrastructure or state file present for the new code that I am using terraform-validator to scan against.

I have created terraform-validator as a Docker image and it is being pulled into the Cloud Build pipeline run from a private Artifact Registry repository.

melinath:
@ranjitk-burwood Hi! We should be able to have someone take a look at this in the next few days. Are you able to run terraform-validator with `--verbosity=debug`? That may provide additional information that helps us assist with debugging.

ndmckinley:
[...]

PR #436, opened 2021-12-14
Changed behavior for 403 on IAM resources 

ScottSuarez:
Changed behavior for 403 on IAM resources to simply skip the asset. 

It's possible the resource has not been created yet so there is no way to differentiate permission from existance. 
closes https://github.com/GoogleCloudPlatform/terraform-validator/issues/393


melinath (approved):
LGTM - would it be possible to add tests?

ScottSuarez:
seems more effort than worth. There isn't existing infrastructure set up to test this functionality, would easily multiply the development time considerably.

Issue #372, opened 2021-11-09
add error levels to stderr output


melinath:
Currently the stderr output doesn't state what level the error is at, which is confusing and makes it look like even info-level issues are blockers. If the output is human-facing i.e. non-structured, we should add back in the error levels to make it easier to parse what's happening.

PR #438, opened 2021-12-14
Add level key to show log level if useStructuredLogging is false

iyabchen:
Resolved https://github.com/GoogleCloudPlatform/terraform-validator/issues/372


melinath (approved):
LGTM.

PR #453, opened 2022-01-06
don't use verbose flag for testing strategy

ScottSuarez:
For our testing strategy lets not use the verbose flag reduce output..

golang prints the output for failed tests anyways.


melinath (approved):
LGTM - this seems like an easy win since the failures will be output regardless!

PR #461, opened 2022-01-11
Log level after gcloud change

iyabchen:
This is to address https://github.com/GoogleCloudPlatform/terraform-validator/issues/372

gcloud need to be changed for structured output schema to accept log level first, else the schema validation complains about the new field. The change is released with version 368.0. 

Now that with the change, we can encode the level key in both structured logging and non structured logging. 

Tested with the latest release of gcloud 368.0 version, and tested with such a command 
```
$ gcloud alpha resource-config validate terraform /home/ciris/test-tf/tfplan.json --policy-library=~/test-tf/policy-library1 --verbosity info
```
- with the default terraform-validator binary (the one downloaded by gcloud)
```
ERROR: Error: [[INFO] Authenticating using configured Google JSON 'access_token'...].
ERROR: Error: [[INFO]   -- Scopes: [https://www.googleapis.com/auth/cloud-platform https://www.googleapis.com/auth/userinfo.email]].
ERROR: Error: [[INFO] Authenticating using configured Google JSON 'access_token'...].
...
```
- with the compiled binary with this change,
```
ERROR: Info: [[INFO] Authenticating using configured Google JSON 'access_token'...].
ERROR: Info: [[INFO]   -- Scopes: [https://www.googleapis.com/auth/cloud-platform https://www.googleapis.com/auth/userinfo.email]].
ERROR: Info: [[INFO] Authenticating using configured Google JSON 'access_token'...].
...
```

It shows with the change, it is able to resolve the log level. The starting `ERROR:` is prefixed by gcloud.

### DIFF
--- a/cmd/logger.go
+++ b/cmd/logger.go
@@ -99,7 +99,7 @@ func newErrorLogger(verbosity string, useStructuredLogging bool, writeSyncer zap
 
 	encoderConfig := zapcore.EncoderConfig{
 		TimeKey:       "timestamp",
-		LevelKey:      "",
+		LevelKey:      "level",
 		MessageKey:    "",
 		StacktraceKey: "",
 		EncodeLevel:   zapcore.LowercaseLevelEncoder,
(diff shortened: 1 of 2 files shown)


ScottSuarez:
Could you explain these changes in more detail?

ScottSuarez (approved):

ScottSuarez:
Is there a way to remove the `ERROR:` from gcloud prefix? This seems like an error on their part.

iyabchen:
@ScottSuarez  I don't think that can be easily removed. Our logger uses field `error_details` to encode, when that field exists, gcloud considers it as an error message, and use `log.error()` to print. The logging configuration is some global setting which sets that head.

PR #468, opened 2022-01-18
Add output option for convert 

iyabchen:
# Background
User perceived tfv stall because it takes long at validation. There proposed a solution to separate convert and validate for gcloud to execute them sequentially. For convert, it will stream the output, and for validate, it uses a progress tracker to show loading animation.
 

# PR
The plan is for convert command to write the result into a file, and validate command to take that file as the input. As a result, the PR does the following. 
- For convert subcommand, add output option. If that output option is not empty, write the converted content into a file. 
- For validate subcommand, support reading converted asset file

### DIFF
--- a/cmd/convert.go
+++ b/cmd/convert.go
@@ -46,6 +46,7 @@ type convertOptions struct {
 	offline           bool
 	rootOptions       *rootOptions
 	readPlannedAssets tfgcv.ReadPlannedAssetsFunc
+	outputPath        string
 }
 
 func newConvertCmd(rootOptions *rootOptions) *cobra.Command {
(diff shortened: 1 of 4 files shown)


melinath (changes requested):
I like this implementation. The only concern I have is reusing the `validate` command's `plan` argument for the CAI asset file, since it means that the argument ends up allowing two types. What would you think about making an explicit option for passing assets instead of a TF plan?

iyabchen:
@melinath  
https://github.com/GoogleCloudPlatform/terraform-validator/blob/11867b988c6add14bfb2ef94d5833b4ad60ea00b/cmd/validate.go#L75
If we create a new option for assets, then we have nothing to fill in for args[0]? And the command looks like this? 
`terraform-validator --asset-path=xxx --policy-path=yyy ...`?

melinath (approved):
@iyabchen That is what I had been thinking, but now that you've written it out it looks like it would just be more confusing for users.

There's a part of me that wants this to be more explicit - for example by making a separate command like `validate-cai` that only has the relevant options (dry-run and policy-path) and only uses structured output and gives an error if the provided file is not assets - but I'm not sure it would add much value at this time, since users are not expected to call this functionality directly and we don't have any third case to take into account.

So, I'm on board with the current implementation.

PR #477, opened 2022-01-21
Standardize user agent between ancestry manager and iam fetching & allow extension

melinath:
This change lets us have a single source of truth for the user agent for TFV API calls, which can be extended similar to the terraform provider's user agent. b/214611094


iyabchen (approved):

melinath:
I went ahead and merged this so that I can test the upstream change prior to merge, which I am now second-guessing as a decision but hopefully it'll be fine 😂

PR #487, opened 2022-01-26
update validator to utilize the new package

ScottSuarez:
This will allow the vendoring package to utilize the local instead of published v4 or tpg library


melinath (changes requested):
If I run `go mod vendor` on this branch, it seems to still pull down tpg - is there something additional that needs to change or that I need to do?

ScottSuarez:
> If I run `go mod vendor` on this branch, it seems to still pull down tpg - is there something additional that needs to change or that I need to do?

if you make a change to the local file you should see it copied downstream to the vendored directory

melinath (approved):

melinath:
It looks like the built binary from this branch ends up being the same size, so the code duplication should be fine.

PR #489, opened 2022-01-27
Fixed release script build var setting & added test to prevent regression

### LINKED ISSUE #488: TFV version output incorrect
I noticed that the tfv version output seems to be `dev` as opposed to semver for some of the latest releases.  I tried a few of the latest 0.12.x.

```
terraform-validator  version
Build version: dev
```

I checked 0.11.0 which worked as expected.
```
./terraform-validator  version
Build version: v0.11.0
```

melinath:
Resolved #488

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -38,7 +38,7 @@ platforms="darwin linux windows"
 skip_platform_arch_pairs=" windows/arm64 "
 
 tar_gz_name=terraform-validator
-ldflags="-X github.com/GoogleCloudPlatform/terraform-validator/tfgcv.buildVersion=v${version}"
+ldflags="-X github.com/GoogleCloudPlatform/terraform-validator/version.buildVersion=v${version}"
 release_bucket=terraform-validator
 
 # Build release versions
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -56,6 +56,20 @@ for platform in ${platforms}; do
 
 		echo "Building ${binary_name} v${version} for platform ${platform} / arch ${arch}..."
 		GO111MODULE=on GOOS=${platform} GOARCH=${arch} CGO_ENABLED=0 go build -ldflags "${ldflags}" -o "${release_dir}/${binary_name}" .
+		if [[ "$platform" == "darwin" && "${arch}" == "amd64" ]]; then
```
iyabchen, line 59: why windows does not need the test?

iyabchen (approved):

release.sh
melinath: If this works for one of the builds, it should work for all of them - and also, this is expected to be run on `darwin amd64` so the other binaries would fail to run.

PR #503, opened 2022-02-02
Moving to a third party tool for diff

prateek2408:
Currently TFV test logs are hard to parse because the diff is output as a giant object. It would be nice to reduce the noice here and assist the developer in pinpointing the exact issues.


ScottSuarez:
Hi sorry for the delay here. Lot has been happening on my end.. I'll look at this a soon as possible

PR #583, opened 2022-03-15
Retry get ancestry with config project when hitting 403 error with the project in tf data

iyabchen:
b/218368396

When project is a random string, get ancestry with that project fails with 403 error. In that case, get ancestry retry with the config.Project.

### DIFF
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -21,10 +21,12 @@ import (
 	"strings"
 	"time"
 
+	"github.com/hashicorp/errwrap"
 	tfjson "github.com/hashicorp/terraform-json"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	provider "github.com/hashicorp/terraform-provider-google/google"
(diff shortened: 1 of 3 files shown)


iyabchen:
> oh - I think I misunderstood what you were talking about. So the situation is:
> 
> 1. project IAM references a random string of numbers (aka a project that doesn't exist)
> 2. trying to fetch the project IAM fails with 403 and correctly pretends that the remote IAM policy is just empty, basically.
> 3. but then when it gets to the augmentation step, it tries to use the project id to get ancestry, which fails again because the project doesn't exist (yet).
> 
> I don't think we want to fall back to the config project in this case, since we know the project id and it doesn't exist - I think only need to fall back if the project id is not known (which is already handled by the getProject() function.)
> 
> But if you disagree or have questions, happy to talk about it!

What is the expected behavior in that case? Just return an error?

iyabchen:
> The expected behavior would be to return an error, yes.

Great! In that case, I will just close this PR.

PR #606, opened 2022-03-24
Refactor ancestry manager to use cloud resource manager API to fetch folder ancestry

iyabchen:
b/211908063 to support folder and organization as the ancestry. folder and organization are not bound to projects in the context. 

The test files will be overwritten by magic-module, so the corresponding PR in magic module is 
in https://github.com/GoogleCloudPlatform/magic-modules/pull/6033. 

- for cai type = Project, if a project does not exist(presuming a new project), it gets 403 from cloud resource manager. In that case, ancestry manager does not return an error, but returns the project itself as the only ancestor. 
- Tests
  - the integration test has to specify the real TEST_ORG_ID, TEST_ANCESTRY, TEST_FOLDER_ID to get it work, since in the online mode, it has no cache, and it tries to fetch a folder's parent
  - have to adjust example_project_update to use a hard coded org id in order to match the violation definition.


melinath (changes requested):
Initial pass - I still need to look at the main change but I wanted to leave you these cleanup items if you want to start acting on them. Overall if there are any parts of this that can be split into separate small PRs and handled first, that would be awesome!

converters/google/convert_test.go
```diff
@@ -480,3 +480,97 @@ func TestTimestampUnmarshalJSON(t *testing.T) {
...
+		wantErr   bool
+	}{
+		{
+			name:      "project",
+			ancestors: []string{"projects/123", "folders/456", "organizations/789"},
+			cai:       &resources.Asset{Type: "cloudresourcemanager.googleapis.com/Project"},
+			want:      "//cloudresourcemanager.googleapis.com/folders/456",
+		},
+		{
+			name:      "project with no other ancestors",
```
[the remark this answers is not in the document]
iyabchen: Yes, based on `google_project`, org_id and folder_id are optional, if the project id is also new, then trying to get this project also fails and cannot construct ancestors.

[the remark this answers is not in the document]
melinath: could you add a comment on this test explaining that?

converters/google/convert_test.go
```diff
@@ -480,3 +480,97 @@ func TestTimestampUnmarshalJSON(t *testing.T) {
...
+			ancestors: []string{"projects/123"},
+			cai:       &resources.Asset{Type: "cloudresourcemanager.googleapis.com/Project"},
+			want:      "//cloudresourcemanager.googleapis.com/projects/123",
+		},
+		{
+			name:      "folder",
```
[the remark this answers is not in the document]
melinath: yes, I think? this would be if the folder is being created with Terraform, and the parent isn't yet known

converters/google/convert_test.go
```diff
@@ -480,3 +480,97 @@ func TestTimestampUnmarshalJSON(t *testing.T) {
...
+				t.Fatalf("converter.assetParent(%v, %v) = %v, want = nil", test.cai, test.ancestors, err)
+			}
+			if got != test.want {
+				t.Errorf("converter.assetParent(%v, %v) = %v, want = %v", test.cai, test.ancestors, got, test.want)
+			}
+		})
+	}
+}
+
+func TestAssetParent_Fail(t *testing.T) {
```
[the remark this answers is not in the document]
melinath: as long as there are test cases to ensure that other types of resources are not allowed, I think that sounds sufficient

converters/google/convert_test.go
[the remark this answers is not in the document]
iyabchen: - folder creation is in `google_folder` resource, and currently there is no converter to support that.
- for folder creation, [no folder_id](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_folder) is in the context, and parent, display_name are required fields in terraform data. In future, if we support `google_folder` resource, more logics need to be adjust. eg. 
  - In this PR, the ancestry manager will use API to fetch `folder` type asset parent. For folder creation, this should not be done. 
- seems folder should always have a parent, for creation, this is defined in parent, for existing folder, it can be fetched from API.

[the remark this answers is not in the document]
melinath: Ah - yeah, this is irrelevant since there's no google_folder resource.

iyabchen:
To get the integration test passing, need to set TEST_ORG_ID=<NUMBER1>, TEST_FOLDER_ID=<NUMBER2>, TEST_ANCESTRY=organization/<NUMBER1>/folder/<NUMBER2> to have the real values.

ancestrymanager/ancestrymanager.go
```diff
@@ -1,189 +1,240 @@
...
-		return path, nil
+func (m *manager) getAncestorsWithCache(key string) ([]string, error) {
+	var ancestors []string
+	cur := key
+	for cur != "" {
+		if cachedAncestors, ok := m.ancestorCache[cur]; ok {
+			ancestors = append(ancestors, cachedAncestors...)
+			break
+		}
+		if strings.HasPrefix(cur, "organizations") {
```
melinath, line 171:
```suggestion
		if strings.HasPrefix(cur, "organizations/") {
```

melinath (changes requested):
took another look, still have a ways to go - it looks like there's some unit test failures that may not be related to the missing env vars?

ancestrymanager/ancestrymanager.go
```diff
@@ -1,189 +1,240 @@
...
-			if !ok || projectID == "" {
-				return nil, false
+func (m *manager) initAncestryCache(entries map[string]string) {
+	for item, ancestry := range entries {
+		if item != "" && ancestry != "" {
+			ancestors, err := parseAncestryPath(ancestry)
+			if err != nil {
+				continue
+			}
+			if !strings.HasPrefix(item, "projects/") && !strings.HasPrefix(item, "folders/") {
```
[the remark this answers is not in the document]
iyabchen: What problems could it introduce? If user provides invalid key, then the cache is invalid, then it will use cloud resource API to fetch ancestors.

ancestrymanager/ancestrymanager.go
iyabchen: Done

ancestrymanager/ancestrymanager.go
```diff
@@ -1,189 +1,240 @@
...
-		},
-		&cloudresourcemanager.Ancestor{
-			ResourceId: &cloudresourcemanager.ResourceId{
-				Type: "organization",
-				Id:   "unknown",
-			},
-		},
-	}, nil
+// New returns AncestryManager that can be used to fetch ancestry information
+// for a project. entries takes project as key and ancestry as value to
```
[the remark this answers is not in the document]
iyabchen: Done

ancestrymanager/ancestrymanager.go
```diff
@@ -1,189 +1,240 @@
...
+	// GetAncestors returns a list of ancestors.
+	GetAncestors(config *resources.Config, tfData resources.TerraformResourceData, cai *resources.Asset) ([]string, error)
 }
 
-// resourceAncestryManager provides common methods for retrieving ancestry from resources
-type resourceAncestryManager struct {
+type manager struct {
+	// The logger.
 	errorLogger *zap.Logger
+	// GCP resource manager service. If this field is nil, online lookups will .
```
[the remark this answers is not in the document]
iyabchen: Done

ancestrymanager/ancestrymanager.go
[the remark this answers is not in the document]
iyabchen: What should be the expected behavior? 
For current code, if it does not start with projects/ or folders/, then it is considered as a pure project name(mostly in that case, we do not need to fix the test cases), so no error will be surface. 
I think it can be less restrictive, to allow it to check whether it check project(s)/ and folder(s)/, but I am not sure whether we should surface if users put "orgs/", "organizations/" etc?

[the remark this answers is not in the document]
melinath: My understanding is that a valid cache id is either:

- `folders/<number>`
- `projects/<number>`
- `projects/<id>` (at some point we might want to maintain a separate project id -> project number cache but that's separate work)

Currently if a user provided a value like `organizations/12345`, my read of the code is that it would be put in the cache as `projects/organizations/12345` - which wouldn't really hurt anything but isn't what the user intended.

A better example might be if a user provided a value like `foldres/12345` (typo). Currently that would end up in the cache as `projects/foldres/12345`, which is not what the user intended and could actually negatively impact validation behavior.

For the expected behavior - it would be great for invalid cache keys to be surfaced to the user as errors so they can be fixed - they're essentially input errors.

None of those cases support a `/` on the right-hand side, so one simple validation could be to make sure that if a `/` is present, the left side is a valid resource type.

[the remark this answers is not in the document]
iyabchen: Done. I added check for both key and value. 
If any of them contains invalid resource type (not projects/, folders/, organizaitons/), init ancestry manager throws an error.

ancestrymanager/ancestrymanager.go
```diff
@@ -1,189 +1,263 @@
...
+	normStr := normalizeAncestry(path)
+	splits := strings.Split(normStr, "/")
+	if len(splits)%2 != 0 {
+		return nil, fmt.Errorf("unexpected format of ancestry path %s", path)
 	}
-	for project, ancestry := range entries {
-		if ancestry != "" {
-			ancestry = fmt.Sprintf("%s/project/%s", ancestry, project)
+	var ancestors []string
+	for i := len(splits) - 1; i >= 0; i = i - 2 {
```
melinath, line 231: not a blocker but it seems like this might be a little easier to read if `i` was the index of the front part and the loop went forward through the ancestry - something like:

```suggestion
	allowedPrefixes := map[string]bool{
		"projects": true,
		"folders": true,
		"organizations": true,
	}
	for i := 0; i < len(splits); i = i + 2 {
		if _, ok := allowedPrefixes[splits[i]]; !ok {
			return nil, fmt.Errorf("invalid ancestry path %s with %s", path, splits[i])
		}
		ancestors = append(ancestors, fmt.Sprintf("%s/%s", splits[i], splits[i+1]))
	}
```

ancestrymanager/ancestrymanager_test.go
```diff
@@ -22,130 +28,787 @@ func newTestResourceManagerClient(opts []option.ClientOption) *cloudresourcemana
...
-					ResourceId: &cloudresourcemanager.ResourceId{
-						Id:   "my-org",
-						Type: "organization",
-					},
-				},
-			},
-			expectedOutput: "organization/my-org/project/my-prj",
-		},
-	}
+type testTFData struct {
```
melinath, line 31: It might be worth using the [FakeResourceData](https://github.com/GoogleCloudPlatform/terraform-validator/blob/5526b84757265f03cd188fb6129844c0849dfa17/converters/google/fake_resource_data.go#L132) struct here to maximize similarity to behavior from tf data created by the converter - does that seem reasonable?

melinath: answering my own question: having testTFData here is better. Allows for compact definition without requiring a provided schema. The behavior might be slightly different for edge cases but probably none of those apply for the cases represented by these tests.

ancestrymanager/ancestryutil.go
```diff
@@ -0,0 +1,96 @@
...
+	if err := validateAncestors(ancestors); err != nil {
+		return "", err
+	}
+	if cai == nil {
+		return "", fmt.Errorf("asset not provided")
+	}
+	switch cai.Type {
+	case "cloudresourcemanager.googleapis.com/Folder":
+		if len(ancestors) < 2 {
+			return "", fmt.Errorf("unexpected ancestors %s", ancestors)
```
melinath, line 21: "unexpected ancestors" sounds like the list of ancestors contains some ancestors that were not expected. Suggestion for here and other places:

```suggestion
			return "", fmt.Errorf("unexpected value for ancestors: %s", ancestors)
```

ancestrymanager/ancestryutil.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"fmt"
+	"strings"
+
+	resources "github.com/GoogleCloudPlatform/terraform-validator/converters/google/resources"
+)
+
+// AssetParent derives a resource's parent from its ancestors.
+func AssetParent(cai *resources.Asset, ancestors []string) (string, error) {
+	if err := validateAncestors(ancestors); err != nil {
```
melinath, line 12: If I understand correctly, the validation here is in case a user calls this with invalid inputs. What would you think about making GetAncestors also return a value for the parent?

I think this is probably fine as-is but it feels like we end up having to do additional input validation because of having it in a separate function.

ancestrymanager/ancestryutil.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	resources "github.com/GoogleCloudPlatform/terraform-validator/converters/google/resources"
+)
+
+// AssetParent derives a resource's parent from its ancestors.
+func AssetParent(cai *resources.Asset, ancestors []string) (string, error) {
```
melinath, line 11: Possibly name with the same convention as GetAncestors?

```suggestion
func GetParent(cai *resources.Asset, ancestors []string) (string, error) {
```

ancestrymanager/ancestryutil.go
melinath: Alternately, this function could call GetAncestors internally rather than taking a list of ancestors - but I think that may feel more complicated.

ancestrymanager/ancestryutil.go
```diff
@@ -0,0 +1,96 @@
...
+		return "", nil
+	case "cloudresourcemanager.googleapis.com/Project":
+		if len(ancestors) < 1 {
+			return "", fmt.Errorf("unexpected ancestors %s", ancestors)
+		}
+		if len(ancestors) > 1 {
+			return fmt.Sprintf("//cloudresourcemanager.googleapis.com/%s", ancestors[1]), nil
+		}
+		// project creation/update
+		return fmt.Sprintf("//cloudresourcemanager.googleapis.com/%s", ancestors[0]), nil
```
melinath, line 37: Does this match the current behavior for projects? if so, we might need to make the handling of folders match this behavior for now - i.e. if the ancestry isn't known, use the folder itself.

utils/utils.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package google
+package utils
```
melinath, line 15: could this go in `ancestrymanager` instead of `utils`?

melinath (changes requested):
Got through the whole PR. overall LGTM - I have a few questions about potential changes in behavior and some small tweaks to request.

ancestrymanager/ancestrymanager.go
```diff
@@ -1,189 +1,263 @@
...
+			ancestors := append([]string{fmt.Sprintf("projects/%s", projectID)}, ret...)
+			return ancestors, nil
 		}
-		return nil, false
+		key = fmt.Sprintf("projects/%s", projectID)
+		ancestors, err := m.getAncestorsWithCache(key)
+		if err != nil {
+			if isGoogleApiErrorWithCode(err, 403) {
+				// new project
+				return []string{key}, nil
```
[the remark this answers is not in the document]
iyabchen: the current behavior for new project seems to be returning 403, convert error. 
example_project_create.tf uses `project_id = "{{.Provider.project}}"`, so I don't think it actually creating a new one.

ancestrymanager/ancestrymanager.go
iyabchen: Done. It had to reverse the sequence at the end though, if i start from 0.
I added the reverse at the end.

ancestrymanager/ancestrymanager_test.go
iyabchen: it causes some circular dependency. 
fake resource data is in converters/google package, and convert.go reference ancestry manager package. If ancestry manager uses fake resource data, then it needs to import fake resource data. 
I move it to another folder/package called tfdata.

ancestrymanager/ancestryutil.go
iyabchen: Done

ancestrymanager/ancestryutil.go
iyabchen: It would require to change the ancestry manager interface in that case. 
Changing the interface: GetAncestors ->   func Ancestors(...) (ancestors []string, parent string, error)

ancestrymanager/ancestryutil.go
iyabchen: The current behavior just put projects/X as parent, no matter what resource it is. X is discovered by getProject, so if the resource does not have a project, then it uses the flag). 
The change is, no matter the project is newly create or not
- if the ancestor list (should at least contain the project itself) has more than itself, then it will use ancestors[1]. 
- if the ancestor list is only having the project itself (in the case project is newly created, but it does not specify folder id or org id), then it will use the project itself as parent, which is ancestors[0].

utils/utils.go
iyabchen: Done

ancestrymanager/ancestryutil.go
iyabchen: the style guide says: Function and method names should not use a Get or get prefix, unless the underlying concept uses the word "get" (eg an HTTP GET)? 
I am going to remove the "Get" in the ancestor manager interface too.

melinath (changes requested):
a couple questions and minor changes - I think this is looking good overall though. Just need to get those env vars set up for you!

ancestrymanager/ancestrymanager.go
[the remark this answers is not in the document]
iyabchen: If the field is empty, terraform plan does not pass.
If the field is referencing another variable, eg. random string, then it will be interpreted as empty, and when it detects cai type is project, and retrieving a project id, then it will return error "required field is not set".

If the field is some new value, that returns 403, blocking it to get ancestors or parent should not block it to be converted, right? 

Do you remember which scenario it might cause a side effect? I think we should add such a test case for that.

ancestrymanager/ancestryutil.go
```diff
@@ -37,35 +36,11 @@ func AssetParent(cai *resources.Asset, ancestors []string) (string, error) {
...
-			s := strings.TrimPrefix(ancestor, "projects/")
-			if len(s) == 0 {
-				return fmt.Errorf("ancestor %v does not have an ID in %v", ancestor, ancestors)
-			}
-		} else {
-			return fmt.Errorf("unexpected type of ancestor %v in %v", ancestor, ancestors)
-		}
-	}
-	return nil
+	return "", fmt.Errorf("unexpected value for ancestors:: %v", ancestors)
```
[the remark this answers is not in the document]
iyabchen: Done.

ancestrymanager/ancestrymanager.go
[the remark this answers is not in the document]
melinath: Ah - I see. This is when the user doesn't have permissions to access the project id that's listed.

We should continue to return the 403 in that case, at least for now.

I agree this is behavior we might want to change - maybe as a follow-up to go/tfv-dont-require-projects - but it would be good to talk through what exactly we want the behavior to be (and this PR is already making a lot of other changes).

melinath:
@iyabchen The env vars should all be set up on the main branch now!

Makefile
```diff
@@ -1,6 +1,7 @@
 build_dir=./bin
 name=terraform-validator
-
+PROJECT_ID=ciris-project
```
melinath, line 3: It looks like these lines might have snuck in by accident?

melinath (approved):
LGTM - just a couple lines to remove from the makefile.

melinath (approved):

PR #607, opened 2022-03-28
Change ancestry from string to map[string]string

iyabchen:
b/211908063
Change function signatures to use map[string]string instead of a pure string for ancestry cache. This is to support folder ID as a key in the ancestry cache, rather than only project ID. 
This PR is to support a larger ancestry manager refactoring.

### DIFF
--- a/cmd/convert.go
+++ b/cmd/convert.go
@@ -93,7 +93,10 @@ func (o *convertOptions) validateArgs(args []string) error {
 
 func (o *convertOptions) run(plan string) error {
 	ctx := context.Background()
-	assets, err := o.readPlannedAssets(ctx, plan, o.project, o.ancestry, o.offline, false, o.rootOptions.errorLogger)
+	ancestryCache := map[string]string{
+		o.project: o.ancestry,
+	}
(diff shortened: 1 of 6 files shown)


cmd/convert.go
```diff
@@ -93,7 +93,10 @@ func (o *convertOptions) validateArgs(args []string) error {
 
 func (o *convertOptions) run(plan string) error {
 	ctx := context.Background()
-	assets, err := o.readPlannedAssets(ctx, plan, o.project, o.ancestry, o.offline, false, o.rootOptions.errorLogger)
+	ancestryCache := map[string]string{
+		o.project: o.ancestry,
```
melinath, line 97: It looks like o.project is expected to be a project id; I think that's probably fine for this PR? For the larger PR will this need to be prefixed with `projects/`?

melinath (approved):
LGTM

cmd/convert.go
iyabchen: Yes, in theory, but I added in a parser in ancestrymanager to add the prefix if it does not exist.

PR #622, opened 2022-04-04
Revert "Change ancestry from string to map[string]string"

melinath:
Reverts GoogleCloudPlatform/terraform-validator#607


melinath:
@iyabchen The failure in magic-modules was:

```
# github.com/GoogleCloudPlatform/terraform-validator/test [github.com/GoogleCloudPlatform/terraform-validator/test.test]
test/read_test.go:141:85: cannot use data.Ancestry (type string) as type map[string]string in argument to tfgcv.ReadPlannedAssets
```

I think the issue is that file is managed by magic-modules, so it needs to be updated there simultaneously.

PR #675, opened 2022-05-03
Clarified gcloud beta terraform vet recommendation and removed duplicate documentation

melinath:
Starting with 0.14.0 we are no longer publishing binaries of the terraform-validator library, since that functionality is now better used via [gcloud beta terraform vet](https://cloud.google.com/docs/terraform/policy_validation).

I recommend hiding whitespace changes when reviewing this, since it includes 0.14.0 updates of the supported resources docs (which are mostly the same except whitespace.)


docs/install.md
```diff
@@ -11,23 +11,29 @@ for example:
...
 chmod 755 terraform-validator
 ```
 
 The full list of releases, with release notes, is available [on Github](https://github.com/GoogleCloudPlatform/terraform-validator/releases).
+
+Binary builds are only available for versions up to 0.13.0. 0.14.0+ are only available via [`gcloud beta terraform vet`](https://cloud.google.com/docs/terraform/policy_validation)
+
+## Disclaimer
+
+This is not an officially supported Google product.
```
ScottSuarez, line 39: Perhaps put this at the top of the page within a banner to warn the user. Then this would be the first thing they see within the page.

README.md
```diff
@@ -4,39 +4,15 @@ A gcloud integration for Terraform Validator is in [Public Preview](https://clou
...
-
-Constraints are designed to be compatible with tools across the deployment
-lifecycle. The same set of constraints that you use with Terraform Validator
-can also be used with any other tool that supports them, either at deploy-time
-or as an audit of deployed resources. These constraints live in your
-organization's repository as the source of truth for your security and
-governance requirements. You can obtain constraints from the
-[Policy Library](./docs/policy_library.md), or
-[build your own constraint templates](https://github.com/GoogleCloudPlatform/policy-library/blob/master/docs/constraint_template_authoring.md).
+Terraform Validator is a library used by [`gcloud beta terraform vet`](https://cloud.google.com/docs/terraform/policy_validation).
```
ScottSuarez, line 7: Would recommend more details about the usecase of the library and `terraform vet`'s functionality. Potential to sell the user on the product here.

ScottSuarez (approved):
Comments on formatting and content but no blockers.

README.md
melinath: Took a stab at this. I don't want to get too much into details because it might cause confusion - lmk what you think.

docs/install.md
melinath: I think having this at the bottom of the page might be a best practice, but I can put it at the top as well; that seems reasonable.

PR #698, opened 2022-05-12
Add an no-op ancestry manager

iyabchen:
Add an no op ancestry manager, so internal code can use this for testing, and no need to implement the ancestry manager interface. The interface will need to be modified.


yukinying:
Any possibliity to add unit tests, here or via some upper caller?

yukinying (approved):
LGTM, comment added above.

iyabchen:
> Any possibliity to add unit tests, here or via some upper caller?

Just added a change to modify an existing test case to use the noop ancestry manager.

melinath (approved):
using noop ancestry manager there seems reasonable to me.

PR #718, opened 2022-05-24
Add user agent parameter to ReadPlannedAssets()

roaks3:
This change allows other libraries to call `ReadPlannedAssets()` with an application-specific user-agent. The terraform-validator convert and validate commands will default to `config-validator-tf/*`.

Note that this also deprecates the usage of `GOOGLE_TERRAFORM_VALIDATOR_USERAGENT_EXTENSION` to append to the user-agent.

### DIFF
--- a/cmd/convert.go
+++ b/cmd/convert.go
@@ -17,9 +17,11 @@ package cmd
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"os"
 
 	"github.com/GoogleCloudPlatform/terraform-validator/tfgcv"
+	"github.com/GoogleCloudPlatform/terraform-validator/version"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"go.uber.org/zap"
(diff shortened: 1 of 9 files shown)


cmd/convert.go
```diff
@@ -96,7 +98,8 @@ func (o *convertOptions) run(plan string) error {
 	ancestryCache := map[string]string{
 		o.project: o.ancestry,
 	}
-	assets, err := o.readPlannedAssets(ctx, plan, o.project, ancestryCache, o.offline, false, o.rootOptions.errorLogger)
+	ua := fmt.Sprintf("config-validator-tf/%s", version.BuildVersion())
```
melinath, line 101: I have a slight preference for `userAgent` over `ua` since it's more explicit... but also, I think golang styles tend towards brevity over clarity. Do you have a preference either way?

melinath (changes requested):
LGTM overall - one question

melinath (approved):
lgtm

PR #719, opened 2022-05-24
Added handling for more test vars to CI/CD

melinath:
This is necessary to pass the test vars from the cloudbuild triggers to the test run.


iyabchen (approved):

melinath:
It looks like we'll need to wait for the substitutions to be defined before these tests can succeed - that's covered in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/1143

PR #725, opened 2022-05-25
Simplified converter error messages and added resource address

melinath:
Goals:

- Reduce technical jargon so that errors are easier to understand
- Add resource address to all error messages so that large plans are easier to debug.
- Don't log anything related to non-google resources
- Reduce level of messages that don't impact normal operations
- Add debug-level warnings if there is a beta resource in the plan

b/225383221

### DIFF
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -173,15 +173,25 @@ type Converter struct {
 func (c *Converter) AddResourceChanges(changes []*tfjson.ResourceChange) error {
 	var createOrUpdateOrNoops []*tfjson.ResourceChange
 	for _, rc := range changes {
-		// skip unknown resources
+		// Silently skip non-google resources
+		if !strings.HasPrefix(rc.Type, "google_") {
+			continue
+		}
+
+		// Warn about google-beta resources
+		if rc.ProviderName == "registry.terraform.io/hashicorp/google-beta" {
+			c.errorLogger.Debug(fmt.Sprintf("%s: resource uses the google-beta provider and may not be convertable", rc.Address))
+		}
+
+		// Skip resources not found in the google GA provider's schema
 		if _, ok := c.schema.ResourcesMap[rc.Type]; !ok {
-			c.errorLogger.Info(fmt.Sprintf("unknown resource: %s", rc.Type))
+			c.errorLogger.Debug(fmt.Sprintf("%s: resource type not found in google GA provider: %s.", rc.Address, rc.Type))
 			continue
 		}
 
 		// Skip unsupported resources
 		if _, ok := c.converters[rc.Type]; !ok {
-			c.errorLogger.Info(fmt.Sprintf("unsupported resource: %s", rc.Type))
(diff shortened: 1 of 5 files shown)


converters/google/convert.go
```diff
@@ -173,33 +173,43 @@ type Converter struct {
...
 	for _, rc := range changes {
-		// skip unknown resources
+		// Silently skip non-google resources
+		if !strings.HasPrefix(rc.Type, "google_") {
+			continue
+		}
+
+		// Warn about google-beta resources
+		if rc.ProviderName == "registry.terraform.io/hashicorp/google-beta" {
+			c.errorLogger.Debug(fmt.Sprintf("%s: resource uses the google-beta provider and may not be convertable", rc.Address))
```
roaks3, line 183: Nit: Not sure if there is precedence with "convertable", but I believe it technically should be "convertible"

converters/google/convert.go
```diff
@@ -240,21 +250,21 @@ func (c *Converter) addDelete(rc *tfjson.ResourceChange) error {
...
-					c.errorLogger.Warn(fmt.Sprintf("%s did not return a value for ID field. Skipping asset fetch.", key))
+					c.errorLogger.Debug(fmt.Sprintf("%s: Unable to fetch and merge remote %s asset due to unset or (known after apply) identity fields on the TF resource.", rc.Address, converted.Type))
 					existingConverterAsset = nil
 				} else if errors.Cause(err) == resources.ErrResourceInaccessible {
-					c.errorLogger.Warn(fmt.Sprintf("%s was not able to be fetched due to not existing or insufficient permission. Skipping asset fetch.", key))
+					c.errorLogger.Warn(fmt.Sprintf("%s: Fetching %s for merge failed due to not existing or insufficient permission.", rc.Address, key))
 					existingConverterAsset = nil
 				} else if err != nil {
-					return errors.Wrap(err, "fetching asset")
+					return fmt.Errorf("fetching remote asset %s: %w", key, err)
```
roaks3, line 259: It seems intentional for a few of the conditional blocks you have, but I'm curious why we are not prefixing these errors with an `rc.Address` like the others

converters/google/convert_test.go
```diff
@@ -26,22 +27,49 @@ import (
...
+type bufferWriteSyncer struct {
+	*bytes.Buffer
+}
+
+func (bws bufferWriteSyncer) Sync() error {
+	return nil
+}
+
+// This returns a logger to allow deterministic testing of the output
+// by omitting the timestamp and calling function.
```
roaks3, line 44: ❤️

roaks3 (approved):
LGTM, nothing blocking, but I added one nit and one question

converters/google/convert_test.go
melinath: This code is pretty similar to the zap.NewExample code - tbh I'm a little unsure it is actually deterministic 😂

converters/google/convert.go
melinath: nope, no precedent. just misspelled it. fixed in 414602ab

Issue #105, opened 2019-11-20
ancestry_path is deprecated; use ancestors


yukinying:
https://github.com/forseti-security/config-validator/blob/87b4ae546420814b2a6766a1b4569278f5e1627e/api/validator.proto#L44 
```
  // Ancestor list as returned by CAI (added sometime around Oct 2019)
  repeated string ancestors = 6;
```

I have also verified that CAI now do not return ancestry_path. Instead it returns ancestors, which is a list of strings instead of a single string. For example,

```
		"ancestors": [
			"folders/12345",
			"organizations/67890"
		]
```

I think this will make the logic on generating ancestry be easier.

melinath:
It looks like config-validator uses ancestry_path internally and converts "ancestors" to a path if it's provided - however, I think this would make the expected structure more clear to users and it would be easier to change now than later.

PR #729, opened 2022-05-30
Add ancestors into Asset

iyabchen:
- https://github.com/GoogleCloudPlatform/terraform-validator/issues/105
- changes required in magic module: https://github.com/GoogleCloudPlatform/magic-modules/pull/6091

Adding ancestors into Asset, and update test comparison logic to derive ancestors from ancestry path for the expected object. Test files will be overwritten by the magic module PR.

### DIFF
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -40,7 +40,6 @@ var ErrDuplicateAsset = errors.New("duplicate asset")
 type Asset struct {
 	Name      string         `json:"name"`
 	Type      string         `json:"asset_type"`
-	Ancestry  string         `json:"ancestry_path"`
 	Resource  *AssetResource `json:"resource,omitempty"`
 	IAMPolicy *IAMPolicy     `json:"iam_policy,omitempty"`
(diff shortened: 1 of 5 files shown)


melinath (changes requested):
I see what you mean about not modifying the test files. Avoiding that (annoying, probably manual) churn makes sense to me 👍 

In the long run we probably will want to update the test data to be consistent with the command output for clarity, but having backwards-compatibility for now will keep this PR simple and let us update tests & the TEST_ANCESTRY env var over time.

The one thing that needs to change: The goal is to replace Asset.Ancestry with Asset.Ancestors, rather than having both present. (Test data should be able to specify either one for backwards-compatibility.)

It could be easiest to have a testAsset that has both fields as an intermediate step? But I'm fine with any solution that achieves that goal.

iyabchen:
Done removing the ancestry path field, and in the validate part, add the SanitizeAncestryPath before ReviewAsset .

melinath (approved):
LGTM - I'm fine with doing this workaround for now.

PR #741, opened 2022-06-02
clean up validate code

iyabchen:
Remove sanitize ancestry path, since it is already fixed in config-validator.


melinath (approved):
LGTM

Issue #536, opened 2022-02-17
Error: converting tfplan to CAI assets: adding resource changes to converter: adding resource create or update augmenting asset: getting resource ancestry for project prj-c-billing-logs: googleapi: Error 403: The caller does not have permission


amendezglobant:
I don't know what account the permission should go on and I don't know what permission it is either, I have tried with the cloudbuild account with project viewer and with the terraform account on the project.
Error:
Error: converting tfplan to CAI assets: adding resource changes to converter: adding resource create or update augmenting asset: getting resource ancestry for project prj-c-billing-logs-0843: googleapi: Error 403: The caller does not have permission

melinath:
The permission you need is `resourcemanager.projects.get` for prj-c-billing-logs, which can be granted with the Project Viewer role. If you already have project viewer account I'd recommend making sure that your authentication is set up correctly: https://github.com/GoogleCloudPlatform/terraform-validator/blob/main/docs/user_guide.md#auth

You can also try running the command with `--verbosity=debug` to get more output, which might contain something useful.

If you're not able to resolve the issue, please let us know what terraform-validator version you're using and anything you discover that seems relevant. Thanks!

melinath:
b/220187178

amendezglobant:
Hello, thanks for the information, in my case I´m executing from cloud build service, with this documentation https://github.com/terraform-google-modules/terraform-example-foundation/tree/master/1-org, in my case I have two service account with permission at level organization Project IAM Admin. 
Service Account for terraform org-terraform@prj-b-seed*****.iam.gserviceaccount.com
Service Account Cloud Build  ***********@cloudbuild.gserviceaccount.com 
Error again: 
Error: converting tfplan to CAI assets: adding resource changes to converter: adding resource create or update augmenting asset: getting resource ancestry for project prj-c-billing-logs-0843: googleapi: Error 403: The caller does not have permission

amendezglobant:
I find the same error here.
https://github.com/terraform-google-modules/terraform-example-foundation/issues/620

melinath:
Resolved by https://github.com/GoogleCloudPlatform/terraform-validator/pull/743

PR #743, opened 2022-06-03
Provide URL along with error message for resourcemanager.projects.get 

prateek2408:
Fix for issue https://github.com/GoogleCloudPlatform/terraform-validator/issues/536 .

Adding an extra message that provides the documentation to fix project permission issues

### DIFF
--- a/ancestrymanager/ancestrymanager.go
+++ b/ancestrymanager/ancestrymanager.go
@@ -210,7 +210,7 @@ func (m *manager) getAncestorsWithCache(key string) ([]string, error) {
 		if err != nil {
 			if isGoogleApiErrorWithCode(err, 403) {
 				helperURL := "https://cloud.google.com/docs/terraform/policy-validation/troubleshooting#ProjectCallerForbidden"
-				return nil, fmt.Errorf("for more info- %v, %v", helperURL, err)
+				return nil, fmt.Errorf("User does not have the correct permissions for project %s. For more info: %s", cur, helperURL)
 			}
 			return nil, err
 		}
--- a/ancestrymanager/ancestrymanager.go
+++ b/ancestrymanager/ancestrymanager.go
@@ -8,7 +8,6 @@ import (
 	"google.golang.org/api/cloudresourcemanager/v3"
 
 	resources "github.com/GoogleCloudPlatform/terraform-validator/converters/google/resources"
-	"github.com/pkg/errors"
 
 	"go.uber.org/zap"
 )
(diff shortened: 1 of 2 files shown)


ancestrymanager/ancestrymanager.go
```diff
@@ -208,7 +209,7 @@ func (m *manager) getAncestorsWithCache(key string) ([]string, error) {
 		}
 		project, err := m.resourceManager.Projects.Get(cur).Do()
 		if err != nil {
-			return nil, err
+			return nil, errors.Wrap(err, "for more info- https://cloud.google.com/docs/terraform/policy-validation/troubleshooting#ProjectCallerForbidden")
```
melinath, line 212: A couple things:

1. This wrapping should only happen if this is a 403 error from the API. You will need to use a function like [isGoogleApiErrorWithCode](https://github.com/GoogleCloudPlatform/terraform-validator/blob/07368d5756cbbb7c522f512a5756340a9d14f5b7/converters/google/resources/utils.go#L107) to check that the error is a Google API error and that it has a 403 error code before adding the extra message. (You'll have to copy the logic of that function into the ancestrymanager package, because it's not exported, but that should be fine.) So: a 403 should add this message, but a 500 or a non-API error should not.
2. Instead of using errors.Wrap, could you use `fmt.Errorf("%w: message", err)`? That's the newer way to handle this and makes it a little easier to understand what the final error message will look like.

melinath (changes requested):
This definitely looks like the right spot! Just a couple changes to request.

ancestrymanager/ancestrymanager.go
prateek2408: Fixed, Comments incorporated. However in terms of sending the error message, right now I have appended the error message and the helper URL.

`fmt.Errorf("for more info- %v, %v", helperURL, err)`

Instead of sending the original error message should we just send the helper URL. Something like this 

`fmt.Errorf("for more info- %v", helperURL)`

That will make the error output a bit neat but less informational

melinath: I would be fine with not wrapping the error, but it might be good to make it a little more informational - for example:

```
fmt.Errorf("Project %s does not exist or user does not have the correct permissions. For more info: %s", projectId, helperURL)
```

prateek2408: Changed the error message as per suggestion.

melinath (approved):
LGTM

PR #749, opened 2022-06-07
Update supported resources with 0.15.0 release

roaks3:
I briefly looked through the commits since our last release, and it didn't look like this resource was explicitly added. For that reason, I'm not sure if we need this.


melinath:
It looks like it was added in https://github.com/GoogleCloudPlatform/terraform-validator/pull/699 - it's a bit confusing because the accesspolicy code already existed - the change was just to add the generated resource to the list in resource_converters.go. We require this to be a manual step for non-IAM resources because we don't have a good way to enforce adding tests right now.

melinath (changes requested):
Rather than updating this table, could you replace it with a link to https://cloud.google.com/docs/terraform/policy-validation/create-cai-constraints#supported_resources? It would still be useful to keep the note about `list-supported-resources`, but it would be great to reword it to clarify that the command is only available for OSS builds, which we are not currently providing.

roaks3:
@melinath That makes sense, I just updated the page with your suggestions. How does the updated version look to you?

melinath (approved):

PR #770, opened 2022-06-15
Check folder change before using v3 API

iyabchen:


melinath (changes requested):

ancestrymanager/ancestrymanager.go
```diff
@@ -170,6 +170,20 @@ func (m *manager) fetchAncestors(config *resources.Config, tfData resources.Terr
 			return ancestors, nil
 		}
 		if folderOK {
+			// Get ancestry from project level first, if folder changed, then proceed
+			// with folder. This is to avoid requiring folder level permission if
+			// there is no folder change.
+			folderIsAncestor := false
+			projectAncestors, _ := m.getAncestorsWithCache(fmt.Sprintf("projects/%s", projectID))
+			for _, ancestor := range projectAncestors {
```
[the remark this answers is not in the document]
iyabchen: I am taking the first route here.

ancestrymanager/ancestrymanager_test.go
```diff
@@ -414,6 +414,162 @@ func TestGetAncestors(t *testing.T) {
...
+					"folder_id":  "folders/bar2",
+				},
+			),
+			asset: &resources.Asset{
+				Type: "cloudresourcemanager.googleapis.com/Project",
+			},
+			want:    []string{"projects/foo", "folders/bar2", "organizations/qux2"},
+			parent:  "//cloudresourcemanager.googleapis.com/folders/bar2",
+			v3Count: 1,
+		},
```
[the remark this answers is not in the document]
iyabchen: Added  for both direction.

ancestrymanager/ancestrymanager_test.go
```diff
@@ -414,6 +414,225 @@ func TestGetAncestors(t *testing.T) {
...
+					{ResourceId: &crmv1.ResourceId{Id: "qux", Type: "organization"}},
+				},
+			},
+			want:    []string{"projects/foo", "folders/bar2", "organizations/qux2"},
+			parent:  "//cloudresourcemanager.googleapis.com/folders/bar2",
+			v3Count: 1,
+			v1Count: 1,
+		},
+		{
+			name: "folder changed from folders/bar1/folders/bar2 to folders/bar2",
```
melinath, line 538:
```suggestion
			name: "project moved from child folder to parent folder",
```

ancestrymanager/ancestrymanager_test.go
```diff
@@ -414,6 +414,225 @@ func TestGetAncestors(t *testing.T) {
...
+					{ResourceId: &crmv1.ResourceId{Id: "bar", Type: "folder"}},
+					{ResourceId: &crmv1.ResourceId{Id: "qux", Type: "organization"}},
+				},
+			},
+			want:    []string{"projects/foo", "folders/bar", "organizations/qux"},
+			parent:  "//cloudresourcemanager.googleapis.com/folders/bar",
+			v1Count: 1,
+		},
+		{
+			name: "folder changed from folders/bar1 to folders/bar2",
```
melinath, line 509:
```suggestion
			name: "project moved from a top-level folder in one org to a top-level folder in a different org",
```

ancestrymanager/ancestrymanager_test.go
```diff
@@ -414,6 +414,225 @@ func TestGetAncestors(t *testing.T) {
...
+					{ResourceId: &crmv1.ResourceId{Id: "bar2", Type: "folder"}},
+					{ResourceId: &crmv1.ResourceId{Id: "qux", Type: "organization"}},
+				},
+			},
+			want:    []string{"projects/foo", "folders/bar2", "organizations/qux"},
+			parent:  "//cloudresourcemanager.googleapis.com/folders/bar2",
+			v1Count: 1,
+		},
+		{
+			name: "folder changed from folders/bar2 to folders/bar1/folders/bar2",
```
melinath, line 564:
```suggestion
			name: "project moved from parent folder to child folder",
```

ancestrymanager/ancestrymanager_test.go
melinath: If I understand this right - the v1Responses provided for this test show the project being inside `folders/bar`, but the resource data shows it inside `folders/bar2` - so the final destination is a parent of the current location.

ancestrymanager/ancestrymanager_test.go
```diff
@@ -414,6 +414,225 @@ func TestGetAncestors(t *testing.T) {
...
+			data: tfdata.NewFakeResourceData(
+				"google_project",
+				p.ResourcesMap["google_project"].Schema,
+				map[string]interface{}{
+					"project_id": "foo",
+					"folder_id":  "folders/bar2",
```
melinath, line 570:
```suggestion
					"folder_id":  "folders/bar",
```

If the folder structure is the same here as in the child -> parent test, then I would expect this to be moving to `folders/bar` (the child folder)

ancestrymanager/ancestrymanager_test.go
```diff
@@ -414,6 +414,225 @@ func TestGetAncestors(t *testing.T) {
...
+				},
+			),
+			asset: &resources.Asset{
+				Type: "cloudresourcemanager.googleapis.com/Project",
+			},
+			v3Responses: map[string]*crmv3.Project{},
```
melinath, line 576:
```suggestion
			v3Responses: map[string]*crmv3.Project{
				"folders/bar2":       {Name: "folders/bar", Parent: "folders/bar2"},
			},
```

melinath (approved):
LGTM apart from a few minor fixes to the names for the new tests & logic for one.

ancestrymanager/ancestrymanager_test.go
iyabchen: Yes, I guess that's what you mentioned by checking the current ancestry, we find out whether there is a change.

ancestrymanager/ancestrymanager_test.go
iyabchen: are we still filling the v3 responses even though they are not accessed?

melinath: In this test (moving from parent -> child) I would expect the flow to be:

1. Resource starts at `organizations/qux/folders/bar2/projects/foo`
2. Resource folder_id is set to `folders/bar` (a child of `folders/bar2`)
3. ancestryManager pre-warms the cache with GetAncestry, which fetches the old ancestry: `organizations/qux/folders/bar2/projects/foo`
4. Because the old folder doesn't match the new one, the ancestryManager falls back to the v3 API to get the ancestry for `folders/bar`
5. After getting the parent for the child folder, it ends up back at `folders/bar2` and uses the cache for the rest.

Does that match your expectations?

ancestrymanager/ancestrymanager_test.go
```diff
@@ -414,6 +414,233 @@ func TestGetAncestors(t *testing.T) {
...
+					{ResourceId: &crmv1.ResourceId{Id: "bar", Type: "folder"}},
+					{ResourceId: &crmv1.ResourceId{Id: "qux", Type: "organization"}},
+				},
+			},
+			want:    []string{"projects/foo", "folders/bar", "organizations/qux"},
+			parent:  "//cloudresourcemanager.googleapis.com/folders/bar",
+			v1Count: 1,
+		},
+		{
+			// project moving from organizations/qux/folders/bar to organizations/qux2/folders/bar2
```
melinath, line 511: 💯  great to have this added information in the comments!

melinath (approved):
LGTM

PR #778, opened 2022-06-22
Support organizations/unknown

iyabchen:
DO_NOT_MERGE
This is just a draft, and requires magic modules change as well.

### DIFF
--- a/ancestrymanager/ancestrymanager.go
+++ b/ancestrymanager/ancestrymanager.go
@@ -116,6 +116,7 @@ func (m *manager) fetchAncestors(config *resources.Config, tfData resources.Terr
 	key := ""
 	orgKey := ""
 	folderKey := ""
+	projectKey := ""
 
 	orgID, orgOK := getOrganizationFromResource(tfData)
 	if orgOK {
@@ -131,6 +132,14 @@ func (m *manager) fetchAncestors(config *resources.Config, tfData resources.Terr
 			folderKey = fmt.Sprintf("folders/%s", folderKey)
 		}
 	}
+	project, _ := getProjectFromResource(tfData, config, *cai, m.errorLogger)
+	if project != "" {
+		projectKey = project
+		if !strings.HasPrefix(projectKey, "projects/") {
+			projectKey = fmt.Sprintf("projects/%s", project)
+		}
+	}
+
 	switch cai.Type {
 	case "cloudresourcemanager.googleapis.com/Folder":
 		if !folderOK {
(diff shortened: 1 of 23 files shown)


ancestrymanager/ancestrymanager.go
```diff
@@ -313,3 +312,25 @@ type NoOpAncestryManager struct{}
 func (*NoOpAncestryManager) Ancestors(config *resources.Config, tfData resources.TerraformResourceData, cai *resources.Asset) ([]string, string, error) {
 	return nil, "", nil
 }
+
+func (m *manager) handleUnknownProject(config *resources.Config, tfData resources.TerraformResourceData, cai *resources.Asset) ([]string, error) {
```
melinath, line 316: It feels like this duplicates some of the logic in fetchAncestors - I wonder if it would be possible to refactor fetchAncestors so that this function isn't required, now that partial ancestries are allowed? For example, maybe fetchAncestors could get the project, folder, and organization from the current resource, and then build the ancestry in the same way for all resources, like:

1. get org, folder, and project from resource
2. If project is set, get project ancestry
   - If folder / org are set, and match the fetched ancestry, return it.
   - Returns 403 errors from the API
3. If folder is set, get folder ancestry
   - Returns 403 errors from the API
4. Combine project + folder / org ancestries to get full resource ancestry
5. If no ancestry could be built, return "organizations/unknown"

What do you think - would something like that work / be simpler? Would we potentially run into issues with resource-specific behavior?

ancestrymanager/ancestrymanager_test.go
```diff
@@ -1043,3 +1043,114 @@ func TestParseAncestryKey_Fail(t *testing.T) {
...
+func TestHandleUnknownProject(t *testing.T) {
+	p := provider.Provider()
+	cases := []struct {
+		name        string
+		data        resources.TerraformResourceData
+		asset       *resources.Asset
+		v1Responses map[string][]*crmv1.Ancestor
+		v3Responses map[string]*crmv3.Project
+		want        []string
+		parent      string
```
melinath, line 1056: Could you prefix the things being tested with `want`? for example:

```suggestion
		wantParent      string
```

converters/google/resources/project.go
```diff
@@ -34,7 +34,7 @@ func GetProjectCaiObject(d TerraformResourceData, config *Config) ([]Asset, erro
 	}
 	name, err := assetName(d, config, linkTmpl)
 	if err != nil {
-		return []Asset{}, err
+		name = fmt.Sprintf("//cloudresourcemanager.googleapis.com/projects/placeholder-%s", randString(8))
```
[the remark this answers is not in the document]
melinath: In the long run, this might need to be fixed at a higher level - the project and billing info both get created from the same resource data, which means that they should technically have the same project ID. But in practice, it probably doesn't matter since they're separate asset types and (presumably) haven't been created yet.

For now - fixing this in assetName should be fine.

testdata/templates/example_project_create_empty_project_id.json
```diff
@@ -0,0 +1,34 @@
...
+    "asset_type": "cloudbilling.googleapis.com/ProjectBillingInfo",
+    "ancestry_path": "{{.Ancestry}}/project/{{.Provider.project}}",
+    "resource": {
+      "version": "v1",
+      "discovery_document_uri": "https://www.googleapis.com/discovery/v1/apis/cloudbilling/v1/rest",
+      "discovery_name": "ProjectBillingInfo",
+      "parent": "//cloudresourcemanager.googleapis.com/projects/{{.Provider.project}}",
+      "data": {
+        "billingAccountName": "billingAccounts/{{.Project.BillingAccountName}}",
+        "name": "projects//billingInfo"
```
melinath, line 13: If the placeholder logic for project id goes into assetName - I think you could use assetName to build the billing account name, which I think would fix this bug.

melinath (changes requested):
overall looking good, have a few questions & change requests.

ancestrymanager/ancestrymanager.go
iyabchen: I changed the code to basically follow this plan, but it mostly apples to project resource only, and not expanding to other resource types.

ancestrymanager/ancestrymanager_test.go
iyabchen: Done.

testdata/templates/example_project_create_empty_project_id.json
iyabchen: asset name and project ID seem to two separate logic.  assetName issue I addressed in file transport.go, and if getting project attempts all failed, I returned a placeholder;  while this billing data it directly used d.Get("project_id") in project.go.

converters/google/resources/project.go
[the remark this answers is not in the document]
iyabchen: I handled this in transport.go, after trying all kinds of getting project failed, it returns a `placeholder-project` string, while terraform data's project id is still an empty string. In ancestry manager, if terraform data project id is empty, then trigger the org/unknown handling.

ancestrymanager/ancestrymanager.go
melinath: 👍  sounds good & looks good

converters/google/resources/project.go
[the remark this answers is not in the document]
melinath: That probably would work - but unfortunately, we can't make that change in transport.go because it's [shared with the Terraform provider](https://github.com/hashicorp/terraform-provider-google/blob/main/google/transport.go). Could it happen in assetName instead?

testdata/templates/example_project_create_empty_project_id.json
melinath: yes, the implementation currently uses `d.Get("project_id")` - but I think that's wrong, since it would also fail if the project_id was supposed to default to the provider project. I think replacing [this line](https://github.com/GoogleCloudPlatform/terraform-validator/blob/f72ddec9686bcc79a994171c9fce556efa9e497e/converters/google/resources/project.go#L139) with `name = assetName("projects/{{project_id_or_project}}/billingInfo")` would handle the fallbacks correctly, including generating a placeholder. (It would just be a little confusing since it's not actually making an asset name.)

I would have a slight preference for making this change if it's simple and just works, but otherwise we could do that work in a separate ticket.

melinath (changes requested):

converters/google/resources/project.go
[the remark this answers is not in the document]
iyabchen: Added a workaround there to temporarily set the config's project, and unset it later, else need to copy the build template function over?

testdata/templates/example_project_create_empty_project_id.json
iyabchen: changed to split the asset name by `/`, and get the project name, and pass the project name to billing object build function

melinath: that seems like a good solution. 👍

converters/google/resources/cai.go
```diff
@@ -99,10 +99,20 @@ type RestoreDefault struct {
 func assetName(d TerraformResourceData, config *Config, linkTmpl string) (string, error) {
 	re := regexp.MustCompile("{{([[:word:]]+)}}")
 
+	// workaround for empty project
+	placeholderSet := false
+	if config.Project == "" {
+		config.Project = "placeholder-project"
```
melinath, line 105: I think it might be better to have a random placeholder here to match the existing logic. The test suite should automatically ignore comparison of placeholders.

```suggestion
		config.Project = fmt.Sprintf("placeholder-%s", randString(8))
```

Would there be any issue doing it that way?

melinath (approved):
LGTM - one potential change

melinath:
/gcbrun - the test failures look unrelated and maybe flakey?

converters/google/resources/cai.go
iyabchen: Done, had to adjust init_test.go to normalize the fields in the resource data to make that work.

melinath: ah, gotcha. That seems like it'll be okay 👍

melinath (approved):
LGTM

melinath (approved):

PR #877, opened 2022-07-26
Updated config-validator dependency

melinath:
Comparison: https://github.com/GoogleCloudPlatform/config-validator/compare/9c7f9463df8c...ec92f4a06e50f8a3cea3d16eed4d1547142cc272

Only impact to TFV users should be that `organizations/unknown` will be supported in match parameters.


iyabchen (approved):

melinath:
merging despite failing integration tests because they're also failing on main - not caused by this PR.

PR #948, opened 2022-09-02
Recover from interface assertion panic in convert func

iyabchen:
b/239699320

### DIFF
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -239,7 +239,7 @@ func (c *Converter) addDelete(rc *tfjson.ResourceChange) error {
 			if errors.Cause(err) == resources.ErrNoConversion {
 				continue
 			}
-			return fmt.Errorf("failed to convert %s: %s", rd.Kind(), err)
+			return err
 		}
 
 		for _, converted := range convertedItems {
@@ -293,7 +293,7 @@ func (c *Converter) addCreateOrUpdateOrNoop(rc *tfjson.ResourceChange) error {
 			if errors.Cause(err) == resources.ErrNoConversion {
 				continue
 			}
-			return fmt.Errorf("failed to convert %s: %s", rd.Kind(), err)
+			return err
 		}
 
 		for _, converted := range convertedAssets {
--- a/converters/google/convert.go
+++ b/converters/google/convert.go
@@ -17,6 +17,7 @@ package google
 import (
 	errorssyslib "errors"
 	"fmt"
+	"runtime/debug"
 	"sort"
 	"strings"
 	"time"
@@ -238,6 +239,7 @@ func (c *Converter) addDelete(rc *tfjson.ResourceChange) error {
 			if errors.Cause(err) == resources.ErrNoConversion {
 				continue
 			}
+			_ = rd.Id()
 			return fmt.Errorf("failed to convert %s: %s", rd.Kind(), err)
 		}
 
@@ -439,6 +441,7 @@ func convertWrapper(conv resources.ResourceConverter, d resources.TerraformResou
 			default:
 				err = fmt.Errorf("unknown panic error: %v", v)
 			}
+			err = fmt.Errorf("%v\n Stack trace: %s", err, debug.Stack())
 		}
 	}()
 	return conv.Convert(d, config)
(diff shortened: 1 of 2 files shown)


melinath:
I think you said you were concerned about not being able to test this - but it looks like your tests are working fine? Could you elaborate?

iyabchen:
> I think you said you were concerned about not being able to test this - but it looks like your tests are working fine? Could you elaborate?

I am not able to test by deleting a field in .tfplan.json, and trigger the interface assertion. for example, for a  google_compute_instance tfplan, I converted it to .tfplan.json first, then in the file, I delete "description" and its value in .tfplan.json. In `compute_instance.go`, the code was doing `d.Get("description").(string)`, so I thought it should get nil and trigger the error, but it did not. 

The way I can trigger the error is, change `d.Get("description").(string)` -> `d.Get("abc").(string)`, that might match what a terraform provider schema change. 

The unit test I did was, it uses a fake converter with a fake convert function, so it is totally not related to any individual resource converter's logic.

melinath:
> The way I can trigger the error is, change `d.Get("description").(string)` -> `d.Get("abc").(string)`, that might match what a terraform provider schema change.

Ah - yeah, I guess that makes sense. If TFV knows the schema type of the field, then it converts nil to the "zero value" for that schema type. (This is the same behavior that TF uses.) https://github.com/GoogleCloudPlatform/terraform-validator/blob/3ef41d069c256419c843e539d2cec36c492cabe7/tfdata/fake_resource_data.go#L77

So the only way to trigger the panic is for the converter to try to access a field that isn't in the schema TFV has - which is what you've done with `d.Get("abc")`.

melinath:
That seems odd, though - if that's the only trigger, users should never really run into this, since we keep the TPG / TFV versions in sync. Possibly they're on an older version?

melinath:
After thinking about this a little more, I'm not entirely sure that I believe that the TPG and TFV versions being out of sync is the source of this error. It might make sense to just catch panics and display them to users including the resource address so that we can get better debug information from users in the future & more accurately determine a root cause.

iyabchen:
> After thinking about this a little more, I'm not entirely sure that I believe that the TPG and TFV versions being out of sync is the source of this error. It might make sense to just catch panics and display them to users including the resource address so that we can get better debug information from users in the future & more accurately determine a root cause.

Updated the code to include stack trace when it recovers from panic, and added the rd.Kind() into the error message.

converters/google/convert.go
```diff
@@ -233,13 +234,12 @@ func (c *Converter) addDelete(rc *tfjson.ResourceChange) error {
...
 		}
-		convertedItems, err := converter.Convert(rd, c.cfg)
-
+		convertedItems, err := convertWrapper(converter, rd, c.cfg)
 		if err != nil {
 			if errors.Cause(err) == resources.ErrNoConversion {
 				continue
 			}
-			return err
+			return fmt.Errorf("failed to convert %s: %s", rd.Kind(), err)
```
melinath, line 242: in terms of letting the user know the address - I think that once the panic is converted to an error, it will already have the address included here: 

- https://github.com/GoogleCloudPlatform/terraform-validator/blob/53973b8b1e9ed2d9b9d41d5dc747bcf40ea3761b/converters/google/convert.go#L203
- https://github.com/GoogleCloudPlatform/terraform-validator/blob/53973b8b1e9ed2d9b9d41d5dc747bcf40ea3761b/converters/google/convert.go#L210

So it should be safe to leave off the Kind, since that is included in the address, and only return the error. That would keep the error message shorter and easier to understand.

melinath: docs for resource address: https://www.terraform.io/internals/json-format#plan-representation

converters/google/convert.go
```diff
@@ -280,20 +280,20 @@ func (c *Converter) addDelete(rc *tfjson.ResourceChange) error {
...
 
 	for _, converter := range c.converters[rd.Kind()] {
-		convertedAssets, err := converter.Convert(rd, c.cfg)
+		convertedAssets, err := convertWrapper(converter, rd, c.cfg)
 		if err != nil {
 			if errors.Cause(err) == resources.ErrNoConversion {
 				continue
 			}
-			return err
+			return fmt.Errorf("failed to convert %s: %s", rd.Kind(), err)
```
melinath, line 296: likewise I think just returning the error should be enough

melinath (changes requested):
A couple minor requests, otherwise LGTM

converters/google/convert.go
iyabchen: Done.

melinath (approved):

PR #990, opened 2022-09-23
Replace errors.Wrap with fmt.Error

iyabchen:
b/234884313 
Code cleanup, replace errors.Wrap with fmt.Error


melinath (changes requested):
The errors should use `%w` rather than `%s` in order to return wrapped errors: https://pkg.go.dev/fmt#Errorf

iyabchen:
> The errors should use `%w` rather than `%s` in order to return wrapped errors: https://pkg.go.dev/fmt#Errorf

Done

melinath (approved):

Issue #1092, opened 2022-11-07
Storage Bucket IAM ancestry inaccurate


jsmilani:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave _+1_ or _me too_ comments; they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment.
* If the issue is assigned to a user, that user is claiming responsibility for the issue.

### Terraform Validator version

terraform-validator: v0.13.0
AND
gcloud terraform tools: v0.8.0

### Affected Resource(s)

* google_storage_bucket_iam_member

### Terraform Plan JSON

```json
{
  "format_version": "1.1",
  "terraform_version": "1.3.4",
  "variables": {},
  "planned_values": {
    "root_module": {
      "resources": [
        {
          "address": "google_storage_bucket.testdata-bucket",
          "mode": "managed",
          "type": "google_storage_bucket",
          "name": "testdata-bucket",
          "provider_name": "registry.terraform.io/hashicorp/google",
          "schema_version": 0,
          "values": {
            "bucket_policy_only": true,
            "cors": [],
            "default_event_based_hold": false,
            "encryption": [],
            "force_destroy": false,
            "id": "bkt-test-data-development-123",
            "labels": {},
            "lifecycle_rule": [],
            "location": "US-CENTRAL1",
            "logging": [],
            "name": "bkt-test-data-development-123",
            "project": "prj-test-123",
            "requester_pays": false,
            "retention_policy": [],
            "self_link": "https://www.googleapis.com/storage/v1/b/bkt-test-data-development-123",
            "storage_class": "STANDARD",
            "uniform_bucket_level_access": true,
            "url": "gs://bkt-test-data-development-123",
            "versioning": [
              {
                "enabled": true
              }
            ],
            "website": []
          },
          "sensitive_values": {
            "cors": [],
            "encryption": [],
            "labels": {},
            "lifecycle_rule": [],
            "logging": [],
            "retention_policy": [],
            "versioning": [
              {}
            ],
            "website": []
          }
        },
        {
          "address": "google_storage_bucket_iam_member.testdata-bucket-test",
          "mode": "managed",
          "type": "google_storage_bucket_iam_member",
          "name": "testdata-bucket-test",
[...]

PR #1128, opened 2022-11-16
Fetch project with storage API for storage bucket

iyabchen:
https://github.com/GoogleCloudPlatform/terraform-validator/issues/1092

Fetch project with storage API if cai Resource Data does not contain project. Falls back to provider config project if fetching from API fails. 

Thought about adding an intergration test, but it requires the bucket exist to make it work, so just added a unit test.


ScottSuarez:
Thanks for the work iris ^^

PR #1398, opened 2023-02-27
Adding parsing support for Custom Org Policy

jashan-g:
Upstream PR: https://github.com/GoogleCloudPlatform/magic-modules/pull/7381


melinath:
@jashan-g this repository is generated from https://github.com/GoogleCloudPlatform/magic-modules so you will need to make a parallel contribution there - but keeping this PR open makes sense since you're also modifying convert.go. (No need to make the parallel PR yet - can wait until the changes are finalized here.) /gcbrun

converters/google/resources/custom_org_policy.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	return CustomOrgPolicy{
+		Name: d.Get("name").(string),
+		Spec: spec,
+	}, nil
+}
+
+func MergeCustomOrgPolicy(existing, incoming Asset) Asset {
+	existing.Resource = incoming.Resource
```
melinath, line 52: a merge function like this is used for Project, as a workaround for the fact that there are sometimes duplicates of data for a particular project that we want to just ignore - but I wouldn't expect us to need it for any other resources. The other place we use merge functions is for IAM, which is because in Terraform, you can specify partial IAM policies (using member or binding resources) which have to be merged with the current policy from the API to find out what the actual end state will be.

I would recommend not having a merge function like this - unless you're sure that this will be necessary.

converters/google/resources/resource_converters.go
```diff
@@ -1,14 +1,14 @@
...
 //
-//     This file is automatically generated by Magic Modules and manual
-//     changes will be clobbered when the file is regenerated.
+//	This file is automatically generated by Magic Modules and manual
+//	changes will be clobbered when the file is regenerated.
 //
-//     Please read more about how to change this file in
-//     .github/CONTRIBUTING.md.
+//	Please read more about how to change this file in
+//	.github/CONTRIBUTING.md.
```
melinath, line 11: These whitespace changes will get overwritten - could you revert them to make the diffs simpler?

converters/google/resources/resource_converters.go
```diff
@@ -250,6 +250,7 @@ func ResourceConverters() map[string][]ResourceConverter {
 		"google_project_iam_member":           {resourceConverterProjectIamMember()},
 		"google_project_iam_custom_role":      {resourceConverterProjectIAMCustomRole()},
 		"google_organization_iam_custom_role": {resourceConverterOrganizationIAMCustomRole()},
+		"google_org_policy_policy":            {resourceConverterCustomOrgPolicy()},
```
melinath, line 253: It would be preferable to match the name of the Terraform resource.

```suggestion
		"google_org_policy_policy":            {resourceConverterOrgPolicyPolicy()},
```

test/cli_test.go
```diff
@@ -100,6 +100,7 @@ func TestCLI(t *testing.T) {
 		{name: "example_compute_target_https_proxy"},
 		{name: "example_compute_target_ssl_proxy"},
 		{name: "example_container_cluster"},
+		{name: "example_custom_org_policy"},
```
melinath, line 103:
```suggestion
		{name: "example_org_policy_policy"},
```

ancestrymanager/ancestrymanager.go
```diff
@@ -324,6 +324,12 @@ func (m *manager) getProjectFromResource(d resources.TerraformResourceData, conf
 		if ok {
 			return res.(string), nil
 		}
+		
+		res, ok = d.GetOk("parent")
+		if ok && strings.Contains(res.(string), "projects/"){
```
melinath, line 329:
```suggestion
		if ok && strings.HasPrefix(res.(string), "projects/"){
```

melinath: & in the other spots.

converters/google/resources/custom_org_policy.go
melinath: Hmm, looking at this some more, it seems like a resource can have multiple CustomOrgPolicies attached. In that case - maybe the merge function should be appending the `incoming.CustomOrgPolicy` list to the `existing.CustomOrgPolicy` list? It could be good to have a test for this case as well.

melinath (changes requested):
I wasn't able to finish reviewing but I've left some comments that I hope will be useful / move this forward.

converters/google/resources/custom_org_policy.go
```diff
@@ -0,0 +1,184 @@
...
+func resourceConverterCustomOrgPolicy() ResourceConverter {
+	return ResourceConverter{
+		Convert:           GetCustomOrgPolicyCaiObject,
+		MergeCreateUpdate: MergeCustomOrgPolicy,
+	}
+}
+
+func GetCustomOrgPolicyCaiObject(d TerraformResourceData, config *Config) ([]Asset, error) {
+
+	assetNamePattern, assetType, err := getAssetNameAndTypeFromParent(d.Get("parent").(string))
```
melinath, line 17: I am pretty sure this should just use `assetNamePattern = "//cloudresourcemanager.googleapis.com/{{parent}}"` in all cases - this is what's causing the diff due to `placeholder-foobar` in the name, because `assetName` is trying to find `{{project_id}}`, `{{folder_id}}`, or `{{organization_id}}` in the terraform data, and it doesn't exist - so it is replacing that variable with a placeholder.

melinath (changes requested):
In addition to the remaining changes above - it looks like the problem is that config-validator doesn't support the CustomOrgPolicy field on Assets - you'll need to also update https://github.com/GoogleCloudPlatform/config-validator/blob/0dfa3040e8a4a0c0f7944e05048a1918913e0970/pkg/api/validator/validator.pb.go#L32 (and then update the dependency on config-validator in this PR.)

melinath (changes requested):
This needs to be updated to include the changes that were made to the ancestry manager.

melinath:
This is blocked by https://github.com/GoogleCloudPlatform/config-validator/pull/198

melinath (commented):
Removing from review queue since it's blocked - you'll need to re-request review when it's unblocked.

melinath (changes requested):
Okay, so it looks like this is failing because the asset never actually gets passed to the validation framework. TFV combines assets together for the conversion process (so that it can get the most complete data about each asset) and then splits them apart again before passing them to the validation framework.

The problem is that the "splitting" code doesn't include splitting out the v2 org policy: https://github.com/GoogleCloudPlatform/terraform-validator/blob/27ebf167208e56d92dd5345779882caca606179f/tfgcv/validate_assets.go#L81

You'll need to adjust `splitAssets` to add handling of custom org policies. I think it should work at that point, but there is a chance there are additional changes that will need to be made to config-validator.

melinath:
Could you also add some tests for org-level and folder-level org policies, just to be safe?

converters/google/resources/org_policy_policy.go
```diff
@@ -0,0 +1,184 @@
...
+
+func resourceConverterOrgPolicyPolicy() ResourceConverter {
+	return ResourceConverter{
+		Convert:           GetOrgPolicyPolicyCaiObject,
+		MergeCreateUpdate: MergeOrgPolicyPolicy,
+	}
+}
+
+func GetOrgPolicyPolicyCaiObject(d TerraformResourceData, config *Config) ([]Asset, error) {
+
```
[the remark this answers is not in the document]
jashan-g: It does require additional changes in config-validator. Raised the PR in https://github.com/GoogleCloudPlatform/config-validator/pull/199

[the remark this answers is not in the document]
melinath: ah, I was only searching for the field names, not the Getters! 💯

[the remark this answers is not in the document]
melinath: @jashan-g it looks like the issue is https://github.com/GoogleCloudPlatform/config-validator/blob/a4226f60974b18f4ed03995b0c9b5463dc39a3b2/pkg/gcptarget/gcptarget.go#L186 - it needs to also check for v2 org policies - something like the following seems to resolve the issue for me locally:

```
foundV2OrgPolicy := false
if asset["v2_org_policies"] != nil {
	foundV2OrgPolicy = true
}
```

You'll also need to add a block to the following resourceTypes check.

It would also be great to add tests to `TestReviewAsset` to check the behavior of at least the v2 org policy case. It would be nice to also get some base cases in there for the various types of resources if you have time, since it doesn't look like we were testing this previously - but just the one case is fine otherwise.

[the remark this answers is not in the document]
jashan-g: Addressed the issue in https://github.com/GoogleCloudPlatform/config-validator/pull/200/files
Also added tests to TestReviewAsset for v2 org policy and other types of resources.

ancestrymanager/ancestrymanager.go
```diff
@@ -331,6 +331,10 @@ func (m *manager) getProjectFromResource(d resources.TerraformResourceData, conf
 		res, ok = d.GetOk("project_id")
 		if ok {
 			return res.(string), nil
+		} 
+		res, ok = d.GetOk("parent")
```
melinath, line 335: I think this is no longer needed because of the changes to the ancestry manager for Folder support. This will fall through to `getProjectFromSchema` which will parse out the parent.

converters/google/convert.go
```diff
@@ -39,11 +39,13 @@ var ErrDuplicateAsset = errors.New("duplicate asset")
...
-	Type      string         `json:"asset_type"`
-	Resource  *AssetResource `json:"resource,omitempty"`
-	IAMPolicy *IAMPolicy     `json:"iam_policy,omitempty"`
-	OrgPolicy []*OrgPolicy   `json:"org_policy,omitempty"`
+	Name            string             `json:"name"`
+	Type            string             `json:"asset_type"`
+	Resource        *AssetResource     `json:"resource,omitempty"`
+	IAMPolicy       *IAMPolicy         `json:"iam_policy,omitempty"`
+	OrgPolicy       []*OrgPolicy       `json:"org_policy,omitempty"`
+	OrgPolicyPolicy []*OrgPolicyPolicy `json:"v2_org_policies,omitempty"`
```
melinath, line 47: Would it make sense to call this V2OrgPolicies to match config-validator? I would generally prefer to be consistent.

melinath (changes requested):
This is looking good to me. A couple minor points below. This would be a good time to update https://github.com/GoogleCloudPlatform/magic-modules/pull/7381 - that PR needs to apply cleanly on top of this one, which means they will both need to be up to date with the latest version of the main branches of each repository.

(Or more specifically: They should include changes from corresponding commits on the main branches of the two repositories. This should be the same as updating to the latest main branch unless you get unlucky.)

ancestrymanager/ancestrymanager.go
jashan-g: This makes sense. I saw the Info log about being unable to retrieve project_id and assumed it to be an error that I have to debug. I have reverted the change.

melinath (approved):
LGTM, just need the MM pr to be in sync: https://github.com/GoogleCloudPlatform/magic-modules/pull/7381

melinath:
Looks like the MM PR applies cleanly on top of this, more or less - minor changes to list ordering and whitespace. I'll fix the merge conflicts & merge them both later today.

PR #1450, opened 2023-03-13
Changes to ancestrymanager logic to parse "parent" field for project, folder and organization asset types.

JashanBITS:
The PR handles ancestryManager module for the following changes:
1. google_org_policy_policy: The parent information resides in {parent} field of the plan file. The existing logic only looks up {folder_id}, {org_id} etc. for the same. The PR also does a lookup on the {parent} field, after matching for assetType.
2. google_folder: Parent maybe an organization or folder, so it does lookup for both.

### DIFF
--- a/ancestrymanager/ancestrymanager.go
+++ b/ancestrymanager/ancestrymanager.go
@@ -146,10 +146,13 @@ func (m *manager) fetchAncestors(config *resources.Config, tfData resources.Terr
 
 	switch cai.Type {
 	case "cloudresourcemanager.googleapis.com/Folder":
-		if !folderOK {
+		if folderOK {
+			key = folderKey
+		} else if orgOK {
+			key = orgKey
+		} else {
 			return []string{"organizations/unknown"}, nil
 		}
-		key = folderKey
 	case "cloudresourcemanager.googleapis.com/Organization":
 		if !orgOK {
 			return nil, fmt.Errorf("organization id not found in terraform data")
--- a/ancestrymanager/ancestrymanager.go
+++ b/ancestrymanager/ancestrymanager.go
@@ -324,6 +324,12 @@ func (m *manager) getProjectFromResource(d resources.TerraformResourceData, conf
 		if ok {
 			return res.(string), nil
 		}
+		
+		res, ok = d.GetOk("parent")
+		if ok && strings.HasPrefix(res.(string), "projects/"){
+			return res.(string), nil
+		}
+
 		// Fall back to project_id if number is not available.
 		res, ok = d.GetOk("project_id")
 		if ok {
--- a/ancestrymanager/ancestrymanager.go
+++ b/ancestrymanager/ancestrymanager.go
@@ -1,372 +1,133 @@
-// Package ancestrymanager provides an interface to query the ancestry information for a resource.
 package ancestrymanager
 
 import (
 	"fmt"
-	"strconv"
 	"strings"
 
-	crmv1 "google.golang.org/api/cloudresourcemanager/v1"
-	crmv3 "google.golang.org/api/cloudresourcemanager/v3"
-	"google.golang.org/api/storage/v1"
-
 	resources "github.com/GoogleCloudPlatform/terraform-validator/converters/google/resources"
-
-	"go.uber.org/zap"
+	"github.com/hashicorp/errwrap"
+	"google.golang.org/api/googleapi"
 )
 
-// AncestryManager is the interface that fetch ancestors for a resource.
-type AncestryManager interface {
-	// Ancestors returns a list of ancestors.
-	Ancestors(config *resources.Config, tfData resources.TerraformResourceData, cai *resources.Asset) ([]string, string, error)
-}
-
-type manager struct {
-	// The logger.
-	errorLogger *zap.Logger
-	// cloud resource manager V3 client. If this field is nil, online lookups will
-	// be disabled.
-	// cloud resource manager V1 client. If this field is nil, online lookups will
-	// be disabled.
-	resourceManagerV3 *crmv3.Service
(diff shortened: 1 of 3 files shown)


shuyama1 (approved):
LGTM! Adding @melinath to give a second pass, since this is my first validator PR review.

melinath (changes requested):
this seems reasonable; could you add some test cases to the unit tests for each file?

ancestrymanager/ancestryutil.go
```diff
@@ -104,6 +112,17 @@ func getFolderFromResource(tfData resources.TerraformResourceData) (string, bool
 	if ok {
 		return folderID.(string), ok
 	}
+	// folder name is store in display_name for google_folder
+	folderID, ok = tfData.GetOk("display_name")
```
melinath, line 116: There are a number of "display_name" fields on different resources that generally hold "human-friendly" names for resources, so we can't rely on it to hold a folder identifier. Even for folder - I am pretty sure the display_name isn't used in the ancestry. The value returned from this function will be used to cache the ancestry of the folder, so we need to be able to rely on it to be in the standard format - which is `folders/XXXXX`.

Looking at the resource - you'll probably need to either:

1. use `folder.id` and check with a regex that it exactly matches the expected format (because every terraform resource has an `id` field
2. use `folder_id` but add the `folders/` prefix back in - because the folder resource [strips it out](https://github.com/hashicorp/terraform-provider-google/blob/c0c2e42de049c5a0c37ba460626ac7ae113c1f0e/google/resource_google_folder.go#L138).

No matter which you choose, you'll need to make sure (with new test cases) that folder will return a valid ancestry even if folder_id is present on the resource with just a number in it. (Currently, this function assumes that `folder_id` contains something like `folders/XXXX` because on most resources, that's correct.)

melinath (changes requested):

ancestrymanager/ancestryutil.go
jashan-g: I have removed the code fetching {display_name} as an ancestor for google_folder. As discussed in offline conversation, ancestry path will be determined by {parent} field when the {folder_id} for google_folder is not yet set.

ancestrymanager/ancestryutil.go
```diff
@@ -91,10 +95,14 @@ func getOrganizationFromResource(tfData resources.TerraformResourceData) (string
...
+	if ok && strings.HasPrefix(orgID.(string), "organizations/") {
+		return orgID.(string), ok
+	}
 	return "", false
 }
 
-// getFolderFromResource reads folder_id or folder field from terraform data.
+// getFolderFromResource reads folder_id, folder, parent field from terraform data.
 func getFolderFromResource(tfData resources.TerraformResourceData) (string, bool) {
 	folderID, ok := tfData.GetOk("folder_id")
```
melinath, line 107: For a folder that has already been created, this function will currently return the value of `folder_id`, which is a string of numbers `XXXX` - but it needs to instead return `folders/XXXX`. However, for most resources, `folder_id` will _already_ be in the format `folders/XXXX`. So, it needs to conditionally add the prefix if it's not already present.

melinath: The test for this could be something like `Google folder with organizations/ as {parent} and folder_id set`

melinath (changes requested):

ancestrymanager/ancestryutil.go
```diff
@@ -91,17 +95,30 @@ func getOrganizationFromResource(tfData resources.TerraformResourceData) (string
...
 	if ok {
 		return folderID.(string), ok
 	}
+
 	folderID, ok = tfData.GetOk("folder")
 	if ok {
+		if strings.HasPrefix(folderID.(string), "folders/") {
```
melinath, line 114: These changes (from https://github.com/GoogleCloudPlatform/terraform-validator/pull/1450/commits/2357e96c4ce5032af3effc54790eee1f5de712c9) should be reverted - no `folder` field exists on the `google_folder` resource. But it turns out the test you wrote is already passing, because it turns out that the ancestry manager already adds a `folders/` prefix in the calling function: https://github.com/GoogleCloudPlatform/terraform-validator/blob/ea566f835725dc2d98c9ad3c09376f75aeea8aff/ancestrymanager/ancestrymanager.go#L132

So, you don't need to add a prefix to the value of `folder_id` either :-)

melinath (changes requested):

ancestrymanager/ancestryutil.go
jashan-g: Reverted the change since it was adding redundant code.

ancestrymanager/ancestrymanager.go
```diff
@@ -324,6 +327,12 @@ func (m *manager) getProjectFromResource(d resources.TerraformResourceData, conf
 		if ok {
 			return res.(string), nil
 		}
+
+		res, ok = d.GetOk("parent")
```
melinath, line 331: Sorry for not noticing this before; I think the reading of the `parent` field for `projects/XXXX` is currently not implemented the way we'd want. The logic here would only be called for Project / ProjectBillingInfo resources, which shouldn't have a `parent` field (so it'll have no effect.)

The logic in `getProjectFromSchema` is what's making this all work out right - technically, it would be better to get rid of `getProjectFromSchema` and include the logic directly in this function, but I am fine with leaving it there for now.

melinath (changes requested):

ancestrymanager/ancestrymanager.go
melinath: So, the change request is to remove this section since it probably has no effect.

melinath (approved):
LGTM!

PR #1484, opened 2023-03-23
Changed terraform provider being used from google to google-beta

dishikat:
Use Case: To add support for org-policy custom-constraints in terraform-validator (which is a non GA resource, hence supported in terraform/google-beta provider) the provider currently being used in terraform-validator is modified from google to google-beta in this PR.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,8 @@ require (
 	github.com/hashicorp/terraform-json v0.14.0
 	github.com/hashicorp/terraform-plugin-framework v1.1.1
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.24.0
-	github.com/hashicorp/terraform-provider-google-beta v1.20.1-0.20230324233649-b8e2c1c9ac72
+	github.com/hashicorp/terraform-provider-google v1.20.1-0.20230328023035-2a66426b3714
+	github.com/hashicorp/terraform-provider-google-beta v1.20.1-0.20230327033815-6a987a080e13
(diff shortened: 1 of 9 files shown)


go.mod
```diff
@@ -115,6 +115,7 @@ require (
 	github.com/hashicorp/terraform-plugin-go v0.14.3 // indirect
 	github.com/hashicorp/terraform-plugin-log v0.7.0 // indirect
 	github.com/hashicorp/terraform-plugin-mux v0.8.0 // indirect
+	github.com/hashicorp/terraform-provider-google-beta v1.20.1-0.20230322214054-60416c59ce67 // indirect
```
melinath, line 118: This should be a direct dependency, and the non-beta provider shouldn't be a dependency any more.

melinath (changes requested):
looks like something went wrong with the merge.

go.mod
dishikat: Okay thanks Stephen for the direction!
Let me have a look

melinath (changes requested):
This will also require a change to https://github.com/GoogleCloudPlatform/magic-modules/blob/8870ccc7b9e1ea8303c52bbf8865e042568e06c0/.ci/containers/downstream-builder/generate_downstream.sh#L126 to use the beta version as well - that's the code that generates TFV, which modifies the go.mod and go.sum files. (Which is why there keep being merge conflicts.)

melinath:
closing this PR as it'll need to happen in a significantly different way

PR #1486, opened 2023-03-24
Fix asset count

snpd25:
Folder convert logic in magic module PR: https://github.com/GoogleCloudPlatform/magic-modules/pull/7246, is adding one more converted asset. Thus fixing the count of converted asset.


melinath (approved):
LGTM - tests pass when combined with https://github.com/GoogleCloudPlatform/magic-modules/pull/7246

License check failures are unrelated.

PR #1491, opened 2023-03-27
Adding support for Org policy custom constraint

dishikat:
Adding support for org policy custom constraint


melinath:
closing this PR - it'll need to be made against https://github.com/GoogleCloudPlatform/terraform-google-conversion instead (but is also if I remember correctly blocked by switching to the beta provider)

PR #1509, opened 2023-03-29
Updated go version to 1.19

melinath:
Fixes failing image build https://pantheon.corp.google.com/cloud-build/builds/cdfbc10b-2459-49e4-961f-44ca89b8f82b;step=0?jsmode=O&mods=logs_tg_staging&project=config-validator Follow-up for https://github.com/GoogleCloudPlatform/magic-modules/pull/7557


melinath:
Looks like the MM upgrade is being reverted but we might as well make the upgrade here anyway so that it doesn't break again in the future.

shuyama1 (approved):

